PR #9, opened 2019-06-28
minor .gitignore and requirement.txt tweaks

soxofaan:


soxofaan:
tests are failing because added `openeo-driver` dependency in requirements.txt is not available yet on  https://artifactory.vgt.vito.be/api/pypi/python-openeo/simple

Issue #12, opened 2019-07-16
Avoid heavy __init__.py


soxofaan:
`openeogeotrellis/__init__.py` is quite heavy: lot of functions are defined, imports lot of implementation details, triggers some spark related initializations.

This causes some issues: e.g. high risk on circular import issues, undesired side effects of importing `openeogeotrellis` make code reuse harder, ...

It would be better to move things to subpackages and limit `__init__` to imports of the core features 

(related to Open-EO/openeo-python-driver#7)

soxofaan:
also related to Open-EO/openeo-python-driver#8

PR #13, opened 2019-07-17
Refactor secondary service management and allow service deletion (geopyspark-driver part)

soxofaan:
This is the geopyspark part of PR Open-EO/openeo-python-driver#10 to enable deletion of secondary services. Includes also a big refactor to address Open-EO/openeo-python-driver#8


soxofaan:
(Note: also addresses #12  by avoiding new code in `__init__`)

PR #25, opened 2019-10-15
EP-3086 add covjson output format support in aggregate_polygon 

soxofaan:
related to PR Open-EO/openeo-python-driver#23

implements CovJSON output format for aggregate_polygon


soxofaan:
FYI: tests are failing because dependency on PR Open-EO/openeo-python-driver#23

PR #56, opened 2020-11-27
Make it possible to run batch jobs on Kubernetes

tcassaert:
This PR allows someone to execute batch jobs if the driver is running on Kubernetes.

Everything is behind a feature flag that you can enable with the environment variable `KUBE = "true"`. 

The flow goes like this:

1. A bucket on S3 gets created if it doesn't exist yet
2. The `job_info['specification']` gets written to a file in the bucket
3. A Kube manifest (jinja template) gets rendered with variables like object store credentials, bucket name, ...
4. The Kube manifests gets applied to Kubernetes
5. A batch job starts on Kubernetes
6. The job specification object gets read from S3
7. The job specification gets parsed and the batch job starts as usual

This is just about the starting of the batch jobs, but it would be nice to have a review of what I already have.


jdries:
Looks good.
The 'KUBE' feature flag could be replaced with a more generic mechanism, more in line with how we configure the backend implementation, but I believe that would be mostly relevant if we also start seeing other things that need to be different on creo. 
So I'll merge it like this now!

PR #64, opened 2021-01-18
EP-3612 (Sentinel-1 Backscatter layer with Orfeo on Creodias)

soxofaan:
Pull request for EP-3612 (Sentinel-1 Backscatter layer with Orfeo on Creodias) for easier self-review and follow up


jdries:
Most importantly: can we move the backscatter class out of layercatalog, into a separate file? Makes it maybe easier to find.
The output is still written to geotiff and read again, did we try the direct numpy array option in the final version of the flow? (Not a big one, having something that works  is more important.)

soxofaan:
> Most importantly: can we move the backscatter class out of layercatalog, into a separate file

yes indeed I didn't anticipate that the beast would grow so big, will move it for easier overview

> The output is still written to geotiff and read again, did we try the direct numpy array option in the final version of the flow? (Not a big one, having something that works is more important.)

yes I tried that, but couldn't get it working with orfeo package on creodias. I think we have to recompile orfeo binaries with numpy support in some way. I'll make a ticket in our JIRA for that

soxofaan:
ticket about numpy loading from orfeo memory instead of doing geotiff file dump+load: EP-3710

soxofaan:
(rebased to resolve conflict with 905f975a88ce12ae730885c9fa816e9e3ff37f4b and force-pushed)

PR #67, opened 2021-03-04
Update to `ProcessGraphVisitor` API change (namespace support)

soxofaan:
related: Open-EO/openeo-python-client#183


jdries:
ok!

PR #69, opened 2021-03-09
Feature/ep 3670

jdries:


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -93,6 +93,12 @@ def filter_temporal(self, start, end) -> 'GeopysparkCubeMetadata':
     def temporal_extent(self) -> tuple:
         return self._temporal_extent
 
+    @property
+    def opensearch_link_titles(self) -> List[str]:
+        """Get opensearch_link_titles from band dimension"""
+        names_with_aliases = zip(self.band_dimension.band_names, self.band_dimension.band_aliases)
```
soxofaan, line 99: you also might add a prefix like `internal:` to the band name aliases in `layercatalog.json` so you can more robustly detect the internal band names (instead of assuming the first alias is the internal name)

jdries: But then we do need to strip it off again right? When exposing externally.
Will merge for now, to be able to use this on Creo.

jdries:
@laxiwuka  There's failing unit tests on Jenkins for this branch, can you have a look?

openeogeotrellis/geopysparkdatacube.py
soxofaan: stripping off is not necessary when exposing externally, I think, 
it might even be better with prefix exposed  as it indicates that it is a special name for internal use, not for end users

PR #70, opened 2021-03-31
Add DEM layer with file-oscars type

JeroenVerstraelen:
* Added COP_DEM_GLO_30M_COG layer to layercatalog.json
* Added the 'file-oscars' layer type which uses the Sentinel2PyramidFactory 
* Currently the file-oscars layer uses a static cell-size of (30,30) but this will later be pulled out of the cube:dimensions field based on the step size and reference system.

### DIFF
--- a/layercatalog.json
+++ b/layercatalog.json
@@ -1306,12 +1306,15 @@
       "t": {"type": "temporal"},
       "bands": {
         "type": "bands",
-        "values": ["Copernicus_DSM_COG_10"]
+        "values": ["DEM"]
       }
     },
     "summaries": {
       "eo:bands": [
-        {"name": "Copernicus_DSM_COG_10"}
+        {
+          "name": "DEM",
+          "aliases": ["Copernicus_DSM_COG_10"]
+        }
       ]
     },
     "_vito": {
--- a/layercatalog.json
+++ b/layercatalog.json
@@ -1272,5 +1272,56 @@
         "native_crs": "EPSG:4326"
       }
     }
+  },
+  {
+    "id": "COP_DEM_GLO_30M_COG",
+    "name": "COP_DEM_GLO_30M_COG",
+    "title": "Copernicus DEM 30M COG",
+    "description": "Copernicus Global 30 meter Digital Elevation Model dataset in COG format.",
+    "license": "proprietary",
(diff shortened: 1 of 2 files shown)


layercatalog.json
```diff
@@ -1272,5 +1272,56 @@
         "native_crs": "EPSG:4326"
       }
     }
+  },
+  {
+    "id": "COP_DEM_GLO_30M_COG",
```
jdries, line 1277: We'll need a better name for openEO.
Can we use 'COPERNICUS_30'? That aligns it with the Sentinelhub name. Details like 'COG' are not relevant anymore when using openEO.

layercatalog.json
```diff
@@ -1272,5 +1272,56 @@
...
+      "spatial": {"bbox": [[-180, -90, 180, 90]]},
+      "temporal": {"interval": [["2010-12-12", null]]}
+    },
+    "cube:dimensions": {
+      "x": {"type": "spatial", "axis": "x", "step": 0.002777777777777778, "reference_system": "EPSG:4326"},
+      "y": {"type": "spatial", "axis": "y", "step": 0.002777777777777778, "reference_system": "EPSG:4326"},
+      "t": {"type": "temporal"},
+      "bands": {
+        "type": "bands",
+        "values": ["Copernicus_DSM_COG_10"]
```
jdries, line 1309: Let's use 'DEM' as band name, and use the aliases system to ensure that oscars gets the correct link title.

jdries: Actually the complex link title in oscars is also questionable :-)

layercatalog.json
JeroenVerstraelen: Done.

layercatalog.json
JeroenVerstraelen: Done.

JeroenVerstraelen:
The pull request does not appear to update with using my latest commits. Even though the commits are clearly shown at:
[https://github.com/Open-EO/openeo-geopyspark-driver/commits/feature/GDD-1273-copernicus-dem-openeo-layer](https://github.com/Open-EO/openeo-geopyspark-driver/commits/feature/GDD-1273-copernicus-dem-openeo-layer)

Issue #72, opened 2021-04-29
GeopysparkDataCube


soxofaan:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/60361fe93049d7f5aa17c4d0aa98619b9a25ce1e/openeogeotrellis/geopysparkdatacube.py#L190-L195

openeo `apply` process should be pixel level (local) transformation, but `reduce_bands` reduces the band dimension, so there is something wrong with the above, right, @jdries ?

jdries:
It's wrong in the sense that our apply can also be used to reduce bands, and it would work.
But in the actual implementation, I think it really comes down to being the same thing in most cases, or at least I don't know a case where this is a problem.
I guess the improvement would be to be more strict, and pass each band individually into the callback?

soxofaan:
hmm indeed, I just experimented with it and it works.
I was assuming that `reduce_bands` expected a reduce "callback" that takes an input array and returns a single value, but apparently it also works with a unary (single value -> single value) callback, in which case it does not really reduce the data, but just maps it.

soxofaan:
will add comment about this under #73

PR #73, opened 2021-04-29
Some fixes for GeopysparkDataCube.apply

soxofaan:
- `apply` with run_udf failed with "'SingleNodeUDFProcessGraphVisitor' object has no attribute 'builder'"
- result in SingleNodeUDFProcessGraphVisitor code path was not returned
- raise exception when apply use case is not covered

related: #72


jdries (approved):
nice, thanks! Especially good to have the better test coverage.

PR #74, opened 2021-05-03
EP-3718 zonal_statistics: remove special case code path (single polygon timeseries)

soxofaan:
A special case code path (single polygon timeseries) seems to go wrong with certain tile cell types, giving wrong mean values.

In the calculation of `mean = sum / count`, the `sum` seems correct, but the `count` did sometimes include all pixels of the tile, instead of only the non-masked out ones.

This PR removes this special code path (polygonal_mean_timeseries) and simplifies the remaining code a bit

(ref: EP-3718)

### DIFF
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -918,8 +918,8 @@ class GeopysparkDataCube(DriverDataCube):
         def insert_timezone(instant):
             return instant.replace(tzinfo=pytz.UTC) if instant.tzinfo is None else instant
 
-        from_vector_file = isinstance(regions, str)
-        multiple_geometries = from_vector_file or isinstance(regions, GeometryCollection)
+        if isinstance(regions, (Polygon, MultiPolygon)):
+            regions = GeometryCollection([regions])
 
         if func in ['histogram', 'sd', 'median']:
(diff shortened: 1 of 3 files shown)


soxofaan:
Some background of what goes wrong in the current implementation

If you create a GeoPySpark `Tile` without explicit `no_data` or a Python `None` (the default), this is converted at the Scala side to a "no nodata" cell type. In the mask operation that happens in `polygonal_mean_timeseries`, a new result tile is created with same no_data settings as the input, and only the pixels inside the geometry are set. So if the original tile has "no nodata", the result tile is initialized with zeros and when you try to count the "without nodata", you actually count all pixels:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/00a0777a693e24ea1e42783417ed9cd548afbd3d/openeogeotrellis/geopysparkdatacube.py#L1034-L1037

soxofaan:
an alternative fix could be to make sure that the layer before the `mask` operation has a real nodata value.
This will probably be a smaller code change than current PR, but maybe we just want to get rid of too many code paths for the `aggregate_spatial` implementation anyway, @jdries ?

jdries (approved):
Had a look, appears fine. I really think reducing code paths is the better option here, we already have too many of them.

PR #95, opened 2021-07-13
EP-3856 Port from `openeo_udf` to `openeo.udf`

### LINKED ISSUE #78: Don't silently skip invalid UDF code
https://github.com/Open-EO/openeo-geopyspark-driver/blob/20e757b1a1cbc1f72e73301839122fc297576f72/openeogeotrellis/run_udf.py#L74-L104

Currently we don't do anything with the UDF data  if there is no valid UDF signature in the UDF code.
This is confusing when developing with UDFs because things seem to be working (you get output in most cases), but it is not obvious the UDF was not applied

soxofaan:
Fixes #78/#93/EP-3856

Port `run_udf` implementation of GeopysparkDataCube from `openeo_udf` to `openeo.udf`:
- get geopyspark UDF handling in line with how it is documented in openeo python client docs
- eliminates dependency on `openeo_udf`
- throw error when there is no matching/valid UDF function in UDF code (instead of silently doing nothing)
- backwards compatible: old style UDFs that use `openeo_udf` annotations are still supported


setup.py
```diff
@@ -37,9 +37,8 @@
     setup_requires=['pytest-runner'],
     tests_require=tests_require,
     install_requires=[
-        'openeo>=0.7.0a1.*',
+        'openeo>=0.8.1a2.*',
         'openeo_driver>=0.10.11a1.*',
-        'openeo_udf>=1.0.0rc3',
```
soxofaan, line 42: we should probably not drop the dependency on `openeo_udf` yet as existing UDF usage still depends on it. A bunch of tests also depend on it

PR #98, opened 2021-08-03
EP-3819: add initial support for udfs with off-heap memory tiles

JeroenVerstraelen:
Add initial support for off-heap UDFs. Currently, normal UDFs will copy all tile data on spark worker nodes from JVM memory to Python and back. When this off-heap implementation is completed/optimized all tile data should reside in shared memory that is accessible by both Java and Python processes.

- Include runtime as a parameter when calling apply_tiles 
- Added PythonOffheap runtime which will execute the UDF using an initial version of offheap memory tiles (org.openeo.geotrellis.udf.Udf.runUserCode)  
- Added openeo_udf and JEP as dependencies 
- Added the path to libjep.so to LD_LIBRARY_PATH as this a runtime requirement for the JEP library (only on worker nodes)

### DIFF
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -512,7 +512,7 @@ class GeopysparkDataCube(DriverDataCube):
             _log.info(f"apply_tiles tilefunction result dims: {result_array.dims}")
             return (key,Tile(result_array.values, geotrellis_tile[1].cell_type,geotrellis_tile[1].no_data_value))
 
-        if runtime == 'PythonOffheap':
+        if runtime == 'Python-Jep':
             def rdd_function(openeo_metadata: GeopysparkCubeMetadata, rdd):
                 band_names = openeo_metadata.band_dimension.band_names
                 jvm = gps.get_spark_context()._jvm
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -487,7 +487,7 @@ class GeopysparkDataCube(DriverDataCube):
         else:
             raise FeatureUnsupportedException(f"reduce_dimension with UDF along dimension {dimension} is not supported")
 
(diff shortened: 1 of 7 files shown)


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -512,12 +512,26 @@ def tilefunction(metadata: Metadata, openeo_metadata: GeopysparkCubeMetadata,
             _log.info(f"apply_tiles tilefunction result dims: {result_array.dims}")
             return (key,Tile(result_array.values, geotrellis_tile[1].cell_type,geotrellis_tile[1].no_data_value))
 
-        def rdd_function(openeo_metadata: GeopysparkCubeMetadata, rdd):
-            numpy_rdd = rdd.convert_data_type(CellType.FLOAT32).to_numpy_rdd().map(
-                log_memory(partial(tilefunction, rdd.layer_metadata, openeo_metadata))
-            )
-            metadata = GeopysparkDataCube._transform_metadata(rdd.layer_metadata, cellType=CellType.FLOAT32)
-            return gps.TiledRasterLayer.from_numpy_rdd(rdd.layer_type, numpy_rdd, metadata)
+        if runtime == 'PythonOffheap':
```
jdries, line 515: Maybe the name of the runtime can point more to the actual environment, like 'Python-Jep', as that's probably what's going to be the most noticable difference for the user.

setup.py
```diff
@@ -65,6 +65,8 @@
         'Bottleneck==1.3.2',
         'python-json-logger',
         'rlguard-lib',
+        'openeo_udf',
```
jdries, line 68: Is this dependency needed? I don't see an import of something from this package.

jdries (changes requested):
Added some minor remarks.
I believe you also need to modify the submission scripts for batch jobs, under openeogeotrellis/deploy.
The actual internally used deploy script can be found in the mep_deploy repo, which will also require this change.

PR #102, opened 2021-08-19
EP-3717: support orbit filtering

laxiwuka:


openeogeotrellis/collections/s1backscatter_orfeo.py
```diff
@@ -190,11 +191,12 @@ def creodias(
 
         # Build RDD of file metadata from Creodias catalog query.
         # TODO openSearchLinkTitles?
-        attributeValues = {
+        attributeValues = extra_properties
```
soxofaan, line 194: We're copying all `metadata_properties` to `attributeValues` here, most of which will/should be not used.
Can't we add a bit of whitelisting of particular attribute keys here? That will make it also more self-documenting

soxofaan:
thanks!

PR #119, opened 2021-11-18
EP-4067/EP-4068: add drop_dimension and dimension_labels to datacube

laxiwuka:


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -312,6 +312,36 @@ def add_dimension(self, name: str, label: str, type: str = None):
...
+        if dimension == 'temporal':
+            pyramid = Pyramid(map(lambda l: l.to_spatial_layer(), self.pyramid.levels))
+            return GeopysparkDataCube(
+                pyramid=pyramid,
+                metadata=self.metadata.drop_dimension(type=dimension)
+            )
+        else:
+            if not len(self.metadata.bands) == 1:
+                raise OpenEOApiException(status_code=400, message=
+                    """Band dimension can only be dropped if there is only 1 band left in the datacube""")
```
soxofaan, line 329: exception should have `code="DimensionLabelCountMismatch"` 

see https://processes.openeo.org/#drop_dimension

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -312,6 +312,36 @@ def add_dimension(self, name: str, label: str, type: str = None):
             metadata=self.metadata.add_dimension(name=name, label=label, type=type)
         )
 
+    def drop_dimension(self, dimension: str = "other"):
+        if dimension not in ['bands', 'temporal']:
```
soxofaan, line 316: as noted in https://github.com/Open-EO/openeo-python-client/pull/257  `drop_dimension` should drop based on dimension name, not dimension type

soxofaan: and related, from https://processes.openeo.org/#drop_dimension : 
> If a dimension with the specified name does not exist, the process fails with a DimensionNotAvailable exception.

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -312,6 +312,36 @@ def add_dimension(self, name: str, label: str, type: str = None):
...
+                )
+
+    def dimension_labels(self, dimension: str = "other"):
+        if dimension == 'bands':
+            return self.metadata.band_names
+        elif dimension == 'temporal':
+            return list(set(map(lambda k: k.instant, self.pyramid.levels[self.pyramid.max_zoom].collect_keys())))
+        else:
+            raise OpenEOApiException(status_code=400, message=
+                """'dimension_labels' is only supported for dimension types 'bands' and 'temporal'.""")
```
soxofaan, line 343: add `code="DimensionNotAvailable"` to the exception (https://processes.openeo.org/#dimension_labels)

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -312,6 +312,36 @@ def add_dimension(self, name: str, label: str, type: str = None):
...
+                return GeopysparkDataCube(
+                    pyramid=self.pyramid,
+                    metadata=self.metadata.drop_dimension(type=dimension)
+                )
+
+    def dimension_labels(self, dimension: str = "other"):
+        if dimension == 'bands':
+            return self.metadata.band_names
+        elif dimension == 'temporal':
+            return list(set(map(lambda k: k.instant, self.pyramid.levels[self.pyramid.max_zoom].collect_keys())))
```
soxofaan, line 340: I guess users  expect sorted temporal labels, just use `sorted` instead of `list`

openeogeotrellis/geopysparkdatacube.py
soxofaan: also, it's probably better to use same parameter name `name` instead of `dimension`

see https://processes.openeo.org/#drop_dimension  and https://github.com/Open-EO/openeo-processes/issues/305

PR #128, opened 2022-02-25
Fix #126: use real user id in batch jobs instead of proxy user id

### LINKED ISSUE #126: `load_result` not working in batch jobs when using OIDC user
originally reported here https://discuss.eodc.eu/t/caching-openeo-results/197/22

using `load_result` fails with something like
```
 openeo_driver.errors.JobNotFoundException: The batch job ‘4cfc3e73-4847-483f-a580-34d765e86f30’ does not exist.
```

after some digging, I found that the job lookup inside a batch job does not work because the batch job uses the proxyuser as user id while the job registry uses (OIDC based) user id

soxofaan:
see #126 

openEO processing in batch jobs needs real user id for `load_result` and UDPs to work properly.
Proxy user id is just necessary when submitting the job.

### DIFF
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -225,23 +225,21 @@ def main(argv: List[str]) -> None:
     logger.info("argv: {a!r}".format(a=argv))
     logger.info("pid {p}; ppid {pp}; cwd {c}".format(p=os.getpid(), pp=os.getppid(), c=os.getcwd()))
 
-    if len(argv) < 6:
-        print("usage: %s "
-              "<job specification input file> <job directory> <results output file name> <user log file name> "
-              "<metadata file name> [api version] [dependencies]" % argv[0],
-              file=sys.stderr)
-        exit(1)
+    if len(argv) != 9:
+        raise Exception(
+            f"usage: {argv[0]} "
+            "<job specification input file> <job directory> <results output file name> <user log file name> "
+            "<metadata file name> <api version> <dependencies> <user id>"
+        )
 
     job_specification_file = argv[1]
     job_dir = Path(argv[2])
     output_file = job_dir / argv[3]
     log_file = job_dir / argv[4]
     metadata_file = job_dir / argv[5]
-    api_version = argv[6] if len(argv) >= 7 else None
-    dependencies = _deserialize_dependencies(argv[7]) if len(argv) >= 8 else {}
-    # TODO: do we still need proxy user id inside the batch job itself?
-    proxy_user_id = argv[8] if len(argv) >= 9 else None
-    user_id = argv[9] if len(argv) >= 10 else None
+    api_version = argv[6]
+    dependencies = _deserialize_dependencies(argv[7])
+    user_id = argv[8]
 
     _create_job_dir(job_dir)
 
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -239,7 +239,9 @@ def main(argv: List[str]) -> None:
     metadata_file = job_dir / argv[5]
     api_version = argv[6] if len(argv) >= 7 else None
     dependencies = _deserialize_dependencies(argv[7]) if len(argv) >= 8 else {}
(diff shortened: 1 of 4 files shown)


soxofaan:
@bossie   do you think this is going to work?

openeogeotrellis/deploy/batch_job.py
```diff
@@ -239,7 +239,9 @@ def main(argv: List[str]) -> None:
     metadata_file = job_dir / argv[5]
     api_version = argv[6] if len(argv) >= 7 else None
     dependencies = _deserialize_dependencies(argv[7]) if len(argv) >= 8 else {}
-    user_id = argv[8] if len(argv) >= 9 else None
+    # TODO: do we still need proxy user id inside the batch job itself?
+    proxy_user_id = argv[8] if len(argv) >= 9 else None
```
bossie, line 243: The proxy_user_id argument can be removed.

Reasoning:

The proxy user ID is only used to submit a Spark job (spark-submit --proxy-user ...) which only happens in two situations:
- explicit start job in the web API (POST '/jobs/<job_id>/results');
- asynchronously when SHub dumps batch process results in an S3 bucket and we can act on them.

So a batch job never submits batch jobs itself.

openeogeotrellis/deploy/submit_batch_job.sh
```diff
@@ -134,4 +134,4 @@ spark-submit \
  --conf spark.yarn.tags=openeo \
  --jars "${extensions}","${backend_assembly}" \
  --name "${jobName}" \
- "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}"
+ "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}" "${userId}"
```
bossie, line 137: "${proxyUser}" argument can be removed.

openeogeotrellis/deploy/submit_batch_job_spark24.sh
```diff
@@ -144,4 +144,4 @@ spark-submit \
  --conf spark.yarn.tags=openeo \
  --jars "${extensions}","${backend_assembly}" \
  --name "${jobName}" \
- "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}"
+ "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}" "${userId}"
```
bossie, line 147: "${proxyUser}" argument can be removed.

openeogeotrellis/deploy/submit_batch_job_spark3.sh
```diff
@@ -165,4 +165,4 @@ spark-submit \
  --conf spark.yarn.tags=openeo \
  --jars "${extensions}","${backend_assembly}" \
  --name "${jobName}" \
- "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}"
+ "${main_py_file}" "$(basename "${processGraphFile}")" "${outputDir}" "${outputFileName}" "${userLogFileName}" "${metadataFileName}" "${apiVersion}" "${dependencies}" "${proxyUser}" "${userId}"
```
bossie, line 168: "${proxyUser}" argument can be removed.

bossie (changes requested):
I would remove the "${proxyUser}" argument to batch_job.py as it is not needed.

soxofaan:
yes indeed, I didn't see `proxyUser` being used in batch_job.py, but I wanted to be sure it was fine to remove/replace it

PR #133, opened 2022-03-21
EP-3778: process for random forest inference

JeroenVerstraelen:


openeogeotrellis/ml_model.py
```diff
@@ -17,9 +23,88 @@ def write_assets(self, directory: str) -> Dict:
...
 
+    def get_model(self):
+        return self._model
+
     def save_ml_model(self, directory: str) -> Dict:
-        return self.write_assets(directory)
\ No newline at end of file
+        return self.write_assets(directory)
+
+    def create_flask_response(self) -> Response:
```
soxofaan, line 36: I'm not sure this method is necessary. Will it be used at the moment?

soxofaan (approved):
Looks fine to merge already

openeogeotrellis/ml_model.py
JeroenVerstraelen: This is only used when fit_class_random_forest is called with a synchronous request, right? Then it won't be used yet.

PR #157, opened 2022-05-11
Fixing some file format parameter schemas

m-mohr:
Missing enum value / missing null types.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -356,10 +356,10 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
                     "gis_data_types": ["raster"],
                     "parameters": {
                         "tile_grid": {
-                            "type": "string",
+                            "type": ["string", "null"],
                             "description": "Specifies the tile grid to use, for batch jobs only. By default, no tile grid is set, and one Geotiff is generated per date. If a tile grid is set, multiple geotiffs are generated per date, as defined by the specified tiling grid.",
-                            "default": "none",
-                            "enum": ["none","wgs84-1degree","utm-100km","utm-20km","utm-10km"]
+                            "default": None,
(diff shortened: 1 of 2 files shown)


openeogeotrellis/backend.py
```diff
@@ -359,7 +359,7 @@ def file_formats(self) -> dict:
                             "type": "string",
                             "description": "Specifies the tile grid to use, for batch jobs only. By default, no tile grid is set, and one Geotiff is generated per date. If a tile grid is set, multiple geotiffs are generated per date, as defined by the specified tiling grid.",
                             "default": "none",
-                            "enum": ["wgs84-1degree","utm-100km","utm-20km","utm-10km"]
+                            "enum": ["none","wgs84-1degree","utm-100km","utm-20km","utm-10km"]
```
soxofaan, line 362: the default should be `None` instead of `"none"` actually.
I'm not sure if it then has to be part of the enum, or the type should be `["string", "null"]`

soxofaan: as far as I could figure out, any string that is not `"*-degree"` or `"utm-*"`, like  `"none"` , will be handled as if it were `"utm-10km"` (the fallback in openeo-geotrellis-extensions: https://github.com/Open-EO/openeo-geotrellis-extensions/blob/19df3be87e0bb4f32294c270fd1aade063270221/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/tile_grid/TileGrid.scala#L64-L85).

and where we extract the "tile_grid" option from the request,  we use the empty string as default:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/64a950f2bce531d995bc4520a13693419e04190d/openeogeotrellis/geopysparkdatacube.py#L1444

I think we should settle with default `None` (or `null` in JSON ) for this parameter in all these places

Issue #232, opened 2022-10-05
k8s: manage job results directly on object storage


jdries:
Current k8s backend downloads results to a local disk, so they're lost upon restart. It makes more sense to keep them on object storage and  stream files from there as needed.
openeo-geopyspark-driver/openeogeotrellis/job_tracker.py:100

Code that returns data to the user will need to be adjusted to read from object storage directly:
openeo-python-driver/openeo_driver/views.py:998

job metadata lookup will need to be specialized for object storage:
openeogeotrellis.backend.GpsBatchJobs.get_results_metadata

jdries:
job_metadata:

`{"geometry": {"type": "Polygon", "coordinates": [[[-3.72094148493, 41.5479759281], [-3.72094148493, 41.72972578481], [-3.47865906078, 41.72972578481], [-3.47865906078, 41.5479759281], [-3.72094148493, 41.5479759281]]]}, "bbox": [-3.72094148493, 41.5479759281, -3.47865906078, 41.72972578481], "area": {"value": 407487294.17255104, "unit": "square meter"}, "start_datetime": "2021-03-15T00:00:00Z", "end_datetime": "2021-10-31T00:00:00Z", "links": [], "assets": {"openEO.tif": {"href": "/batch_jobs/0146abc4-cf51-46da-9f4c-853781c563aa/openEO.tif", "type": "image/tiff; application=geotiff", "roles": ["data"], "bands": [], "nodata": NaN, "datetime": null}}, "epsg": 32630, "instruments": [], "processing:facility": "VITO - SPARK", "processing:software": "openeo-geotrellis-0.6.3a1", "unique_process_ids": ["array_modify", "arccos", "power", "subtract", "mask_scl_dilation", "save_result", "median", "predict_catboost", "divide", "aggregate_temporal", "reduce_dimension", "load_ml_model", "filter_bands", "apply_dimension", "resample_cube_spatial", "linear_scale_range", "sd", "sar_backscatter", "array_create", "rename_labels", "filter_bbox", "load_collection", "add", "sqrt", "if", "multiply", "mean", "merge_cubes", "array_interpolate_linear", "array_concat", "array_element", "clip", "quantiles", "apply", "is_nodata", "log"]}`

soxofaan:
This issue was closed (indirectly?) by merging Open-EO/openeo-python-driver#157, which is not intended I think, as #280 is still open.

JohanKJSchreurs:
[...]

PR #280, opened 2022-12-01
Issue232 k8s manage job results directly on object storage

JohanKJSchreurs:
# Implement Issue #232, k8s manage job results directly on object storage

These are the modifications on the GeoPySpark driver side to implement issue [Open-EO/openeo-geopyspark-driver/ #232](https://github.com/Open-EO/openeo-geopyspark-driver/issues/232)

This issue requires changes in both the openeo-python-driver and the openeo-geopyspark-driver.


soxofaan:
- [ ] in https://github.com/Open-EO/openeo-python-driver/commit/292f6ab080c1f5a1d41d6b5f5729cd63303f3f1b  I bumped openeo_driver version to 0.28.0,  so also bump it here https://github.com/Open-EO/openeo-geopyspark-driver/blob/168bf1fdb9b776c91ec667d3e6816f5111912bd5/setup.py#L49

openeogeotrellis/backend.py
```diff
@@ -1899,46 +1901,58 @@ def get_results(self, job_id: str, user_id: str) -> Dict[str, dict]:
...
-            }
-
-        for file_name in os.listdir(job_dir):
-            if file_name.endswith("_metadata.json") and file_name != JOB_METADATA_FILENAME:
-                results_dict[file_name] = {
+        # Updating the asset info based on what is on the file system was a solution for some old edge cases.
+        # For a kuberenetes deployment we can skip this, the edge cases should not happen here.
+        # On Kubernetes the result files with be put in the object storage instead of the container,
+        # because the k8s pods can be replaced at any moment and then the files would be gone.
+        if not ConfigParams().is_kube_deploy:
```
soxofaan, line 1908: IMO `is_kube_deploy` is an anti-pattern (see #283), can we avoid introducing it here?

openeogeotrellis/backend.py
```diff
@@ -1899,46 +1901,58 @@ def get_results(self, job_id: str, user_id: str) -> Dict[str, dict]:
...
-            results_dict['profile_dumps.tar.gz'] = {
-                "asset": True,
-                "output_dir": str(job_dir),
-                "type": "application/gzip"
-            }
-
-        for file_name in os.listdir(job_dir):
-            if file_name.endswith("_metadata.json") and file_name != JOB_METADATA_FILENAME:
-                results_dict[file_name] = {
+        # Updating the asset info based on what is on the file system was a solution for some old edge cases.
```
soxofaan, line 1904: "old edge cases" sounds like a legacy, obsolete system. But isn't this "non-kubernetes" way the standard way on the VITO YARN deploy, which is all but obsolete

openeogeotrellis/backend.py
JohanKJSchreurs: I have removed the use of is_kube_deploy in the new code, in favor of a more specific setting: use_object_storage.

For now it is still uses the value of is_kube_deploy in reality, but at least it can now be split up, and I was able to add test coverage to test downloading with use_object_storage set to True/on and False/off.
It is just a question of deciding how exactly we want to configure use_object_storage. (For example, use a new environment variable, or set it True only if SWIFT_BUCKET has a value, etc.)

openeogeotrellis/backend.py
JohanKJSchreurs: This comment turned out to be inaccurate, it is still in use quite a bit. 

Moved away from using is_kube_deploy here.

For the if right below, it was more appropriate to check if the job directory exists. 
https://github.com/Open-EO/openeo-geopyspark-driver/blob/1c2d0be62d822ba4e383009ee86b30ea529e62e0/openeogeotrellis/backend.py#L1921-L1923

And in the for at the bottom, we check with ConfigParams().use_object_storage to see if we should set the output_dir here, or leave it alone because it should have a value already, pointing to S3.
https://github.com/Open-EO/openeo-geopyspark-driver/blob/1c2d0be62d822ba4e383009ee86b30ea529e62e0/openeogeotrellis/backend.py#L1961-L1963

setup.py
```diff
@@ -46,7 +47,7 @@
     tests_require=tests_require,
     install_requires=[
         'openeo>=0.13.0a2.*',
-        'openeo_driver>=0.27.0.dev',
+        'openeo_driver>=0.28.0a1.dev',
```
soxofaan, line 50: FYI: this dep has been bumped to 0.28.3a1 in master already (1d72b14e1fd9e87213526b04559c8c775c003ccf)

tests/test_views.py
```diff
@@ -278,6 +284,49 @@ def test_collections_s2_radiometry(self, api):
...
+
+@pytest.fixture(scope='function')
+def aws_credentials():
+    """Mocked AWS Credentials for moto."""
+    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
+    os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
+    os.environ['AWS_SECURITY_TOKEN'] = 'testing'
+    os.environ['AWS_SESSION_TOKEN'] = 'testing'
+    os.environ['AWS_DEFAULT_REGION'] = TEST_AWS_REGION_NAME
+    os.environ['AWS_REGION'] = TEST_AWS_REGION_NAME
```
soxofaan, line 297: this fixture does not reset on exit. Can we use that monckeypath thing here too?

setup.py
JohanKJSchreurs: I just noticed that upgrading to openeo_driver 0.28.0a1.* brakes a whole lot of tests, 

for example:
```
ERROR tests/test_views.py::TestBatchJobs::test_cancel_job[0.4.0] - AttributeError: type object 'ElasticJobRegistry' has no attribute '_log'
ERROR tests/test_views.py::TestBatchJobs::test_cancel_job[1.0.0] - AttributeError: type object 'ElasticJobRegistry' has no attribute '_log'
```

JohanKJSchreurs: Have merged master into the issues's feature branch. Should be fixed now.

PR #303, opened 2023-01-10
Fix get-jars for windows.

EmileSonneveld:


scripts/get-jars.py
```diff
@@ -68,8 +69,9 @@ def main():
         download_jar(jar_dir, url=url, force=force_download)
 
     logger.info(f"Listing of {jar_dir}:")
-    # TODO: Avoid `ls` subprocess and make this work on Windows too?
-    subprocess.call(["ls", "-al", str(jar_dir)])
+    for f in os.listdir(jar_dir):
+        target = jar_dir / Path(str(f))
+        logger.info(f"{f} ({target.stat().st_size} bytes).")
```
soxofaan, line 74: even shorter without `os`:

```
for f in jar_dir.iterdir():
    logger.info(f"{f.stat().st_size:16d} {f}")
```

EmileSonneveld: nice,.used!

Issue #236, opened 2022-10-08
Ingest jobs and job results into ES


jdries:
Set up synchronization of jobs and job results with ES.
This allows us to build dashboards that give more insight into job specific metrics.

soxofaan:
Closed duplicate #246  I just created

> Reference story for moving from Zookeeper to Elastic Search for batch job metadata.
> Zookeeper was fine as initial POC, but scalability and flexibility (e.g. build an admin UI on top of it) is becoming a problem.
We decided to migrate to an Eleastic Search based solution

also related:
- #245

soxofaan:
internal ticket/ref: FD-5452

soxofaan:
closing as duplicate of https://github.com/Open-EO/openeo-python-driver/issues/153

soxofaan:
Change of plan.
Current approach is to refactor the current `JobTracker` implementation in `job_tracker.py`, but it turns out that it's hard to verify that things keep working on real resources (yarn, k8s, zk, ES, ...), because deploying/running the new version (in a feature branch) in these different contexts is not trivial.  Merging the related PR would also mean a hard switch to the new implementation for all targets (YARN deploy and K8s deploy), with no clean way to roll back (PR currently involved a lot of merge conflicts). This is risky if the new implementation is poorly verified.
 
New plan is to keep original `job_tracker.py` untouched, and just add a new implementation in, say, `job_tracker_v2.py`. Because of how the job trackers work it should be fine to run the original jobtracker concurrently with the v2 version. This gives more room to properly finetune the new implementation

bossie:
[...]

PR #312, opened 2023-01-17
ZkJobRegistry: automatically do `mark_done` (on `set_status`)

soxofaan:
Shrinks public API to simplify abstraction

related to #236

### DIFF
--- a/openeogeotrellis/job_registry.py
+++ b/openeogeotrellis/job_registry.py
@@ -133,6 +133,10 @@ class ZkJobRegistry:
         self.patch(job_id, user_id, status=status, updated=rfc3339.datetime(datetime.utcnow()))
         _log.debug("batch job {j} -> {s}".format(j=job_id, s=status))
 
+        # TODO: move this "mark_done from status" logic to `patch` method?
+        if status in {JOB_STATUS.FINISHED, JOB_STATUS.ERROR, JOB_STATUS.CANCELED}:
+            self._mark_done(job_id=job_id, user_id=user_id)
+
     def set_dependency_status(self, job_id: str, user_id: str, dependency_status: str) -> None:
         self.patch(job_id, user_id, dependency_status=dependency_status)
         _log.debug("batch job {j} dependency -> {s}".format(j=job_id, s=dependency_status))
@@ -158,9 +162,8 @@ class ZkJobRegistry:
 
         self._update({**job_info, **kwargs}, version)
 
-    def mark_done(self, job_id: str, user_id: str) -> None:
+    def _mark_done(self, job_id: str, user_id: str) -> None:
         """Marks a job as done (not to be tracked anymore)."""
-        # TODO: do this automatically from `set_status` with "terminal" status ("finished", "error", ...)
 
         # FIXME: can be done in a transaction
         job_info, version = self._read(job_id, user_id)
@@ -173,6 +176,10 @@ class ZkJobRegistry:
             pass
 
         self._zk.delete(source, version)
+
+        _log.debug(
+            f"marked {job_id=} as done", extra={"job_id": job_id, "user_id": user_id}
+        )
 
     def mark_ongoing(self, job_id: str, user_id: str) -> None:
         """Marks as job as ongoing (to be tracked)."""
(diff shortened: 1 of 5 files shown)


soxofaan:
@bossie   do you think this is going to work?

bossie:
Intuitively, yes this should work because splitting up jobs in ongoing/done is just an optimization so the ZK JobTracker doesn't have to look into all jobs for their status and can just keep an eye on those in the ongoing tree.

This is different from e.g. RDBMS and ES where you can just filter rows/documents by their status.

Ultimately, a job's actual status (created, queued etc) and its ZK subtree have to be in sync.

openeogeotrellis/job_tracker.py
```diff
@@ -131,8 +131,8 @@ def update_statuses(self) -> None:
                                         result_metadata["usage"] = usage
                                     registry.patch(job_id, user_id, **result_metadata)
 
-                                    registry.mark_done(job_id, user_id)
-                                    _log.info("marked %s as done" % job_id, extra={'job_id': job_id})
+                                    # TODO: this `set_status` is just here to trigger the "mark_done from status" logic. Can this be automated further?
+                                    registry.set_status(job_id, user_id, new_status)
```
bossie, line 135: There's also a status change a few lines earlier (with `patch`) so this doesn't feel just right/needs more work.

openeogeotrellis/job_tracker.py
```diff
@@ -181,7 +181,8 @@ def update_statuses(self) -> None:
                                         async_task.schedule_delete_batch_process_dependency_sources(
                                             job_id, user_id, dependency_sources)
 
-                                    registry.mark_done(job_id, user_id)
+                                    # TODO: this `set_status` is just here to trigger the "mark_done from status" logic. Can this be automated further?
+                                    registry.set_status(job_id, user_id, new_status)
```
bossie, line 185: Same here, there's a status change a few lines earlier as well.

openeogeotrellis/job_tracker.py
soxofaan: indeed, the status change above is done with `patch` instead of `set_status` 
and the former does not support this "auto-mark-done" feature (hence the TODO note).

The problem with adding auto-mark-done directly to `patch`, is that it changes the zookeeper path of the batch job which potentially messes up subsequent job_registry operations (e.g. `remove_dependencies` stuff following on `patch`). I wanted to keep this first refactor step simple

openeogeotrellis/job_tracker.py
soxofaan: as mentioned earlier, there is a `remove_dependencies` following on the `patch`, so changing the zookeeper path of a job indirectly from `patch` seemed like risky to include at this point.

but certainly to reconsider as a next step

openeogeotrellis/job_registry.py
```diff
@@ -174,6 +177,10 @@ def mark_done(self, job_id: str, user_id: str) -> None:
 
         self._zk.delete(source, version)
 
+        _log.debug(
+            f"marked {job_id=} as done", extra={"job_id": job_id, "user_id": user_id}
+        )
```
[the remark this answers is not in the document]
soxofaan: aha that is interesting

my attempt was to generalize this log:
 https://github.com/Open-EO/openeo-geopyspark-driver/blob/68bf71e2991657145cd76e3f8ec27700642bb630/openeogeotrellis/job_tracker.py#L135

but apparently it has a side effect.

In what sense will things be messed up? that too much or wrong "marked done" events will be captured, or stopped being captured? Or is is something about the log format and the "extra" fields?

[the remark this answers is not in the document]
bossie: I was thinking along the lines of: that processor might pick up those extra logs and downstream processors might choke on them because they're missing a `unique_process_ids` extra. But yeah, this is to be looked into.

openeogeotrellis/job_tracker.py
bossie: Yes, I believe most/all of those mutators only work on "ongoing" nodes so you want to be careful to not move them to the "done" branch too early.

But as a first step, this is fine by me.

soxofaan:
ok I decided to keep it simpler than this PR  and just do 86d93b4002a98ea33233f094006db59c8a86d263
Just eliminates `mark_done` when immediately after `set_status`.
Should not introduce changed behavior or logic (except maybe for covering `mark_done` calls that were previously forgotten)

PR #313, opened 2023-01-17
ZkJobregistry: automatically do `ensure_paths`

soxofaan:
Shrinks public API to simplify abstraction

related to #236

### DIFF
--- a/openeogeotrellis/deploy/kube.py
+++ b/openeogeotrellis/deploy/kube.py
@@ -35,15 +35,9 @@ def main():
     log.info("starting spark context")
     SparkContext.getOrCreate()
 
-    def setup_batch_jobs():
-        if not ConfigParams().is_ci_context:
-            with ZkJobRegistry() as job_registry:
-                job_registry.ensure_paths()
-
     def on_started():
         app.logger.setLevel('DEBUG')
         deploy.load_custom_processes()
-        setup_batch_jobs()
 
     from openeogeotrellis.backend import GeoPySparkBackendImplementation
     app = build_app(backend_implementation=GeoPySparkBackendImplementation())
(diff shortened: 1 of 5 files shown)


soxofaan:
@bossie  do you think this is going to work?

bossie (commented):
It'll work but I'm not a fan of this approach because it's putting what should be a one-time operation into every instantiation of `ZkJobRegistry`.

At the moment it's done @ startup time because it's convenient if we spin up an entirely new environment and we can just assume ZK is in a usable state, but that's very rare.

I'd prefer just removing the method altogether and making sure that ZK is OpenEO-ready before we try and run it in a new environment (it's just a couple of paths).

soxofaan:
yes indeed, that was also my concern. But my idea (as a next step) was to instantiate just one `ZkJobRegistry` instance at startup and reuse that where necessary, instead of instantiating multiple of them on the fly in multiple places.

That would also simplify injecting alternative job registry implementations (e.g. dummies for testing, or ES based ones).

bossie:
Ok, just call `ensure_paths` on your one instance then:

```python
# at startup
zk_job_registry = ZkJobRegistry()
zk_job_registry.ensure_paths()

# further on, at runtime
def f(job_registry):
  job_registry.set_status(...)  # can be any implementation

```
I'm probably missing something but you should just omit `ensure_paths` from your abstraction, it's ZK specific after all.

soxofaan:
> Ok, just call ensure_paths on your one instance then:

good idea, that indeed simplifies this PR

soxofaan:
I'm actually going to just close this PR as I don't have to change any existing/proven code paths. I just need to play in the new code

PR #316, opened 2023-01-20
Open-EO/openeo-python-driver#160 Skip log lines without a loglevel

JohanKJSchreurs:
This implements https://github.com/Open-EO/openeo-python-driver/issues/160


jdries:
looks good for me

bossie (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ https://openeo.org/documentation/1.0/developers/api/reference.html#operation/con
 If needed, feature flags are used to allow testing unstable features in development/production,
 without compromising stable operations.
 
+## Unreleased
```
soxofaan, line 13: Note: we  don't have a release procedure that takes this changelog into account (like we do in python client).
Instead of `## Unreleased ... ### Changed`, just add a header with (approximate merge) date and current version (see `_version.py`). 
Optionally bump the version in that file within the same commit

soxofaan (approved):
Apart from minor changelog thing, looks good

JohanKJSchreurs:
The CI / Jenkins reports that one unit test fails, but I'm not able to reproduce it locally.

tests/test_download.py::TestDownload::test_download_geotiff_colormap

EmileSonneveld:
The Jenkins error seems indeed related to what I changed. Hexadecimal colors in `cleanDoubleColorMap` are not parsed correctly.
I'll quickly fix and extend the unit test.

EmileSonneveld:
Fixed in dev: https://github.com/Open-EO/openeo-geotrellis-extensions/commit/83f27cd2b15b0f9208287e158d4d1bd286e3e508
@jdries , My local test used color values <16 and so this error was not triggered. I reset my local jars to the normal ones.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ https://openeo.org/documentation/1.0/developers/api/reference.html#operation/con
 If needed, feature flags are used to allow testing unstable features in development/production,
 without compromising stable operations.
 
+## 2023-01-20 (0.6.3a1)
```
soxofaan, line 13: current version is 0.6.7a1:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/125a66f9fc82ebe0812c3bf402b9ab2083e9e944/openeogeotrellis/_version.py#L1

JohanKJSchreurs: I overlooked that, sorry. Now corrected.

JohanKJSchreurs:
Just a small follow up note below, but I know Emile is already on it and that this error is apparently not caused by changes in this PR, so I'm not worried about it anymore.

> colormap was worked on by @EmileSonneveld , did you update your local jars?

That was indeed why I did not get the same error.

I rebuilt my Docker container and manually re-ran the script to download the jars. Now I can reproduce this error.
I think the get_jars.py script skips jar files that are already present so I will have to tune my dockerfile a bit more or add my jars folder to the `.dockerignore` file.

EmileSonneveld:
Meanwhile my fix went trough. ( 10min ago) A new build here should not have this issue anymore

JohanKJSchreurs:
> Meanwhile my fix went trough. ( 10min ago) A new build here should not have this issue anymore

Great. Thanks Emile!

Issue #307, opened 2023-01-12
improve error message when log collection fails


jdries:
our backend returned this at some point when trying to retrieve logs for a batch job:

```
[
  {
    "id": "-1",
    "code": "Internal",
    "level": "error",
    "message": "Log collection failed: ConnectionTimeout('TIMEOUT', \"HTTPSConnectionPool(host='our server', port=443): Read timed out. (read timeout=60)\", ReadTimeoutError(\"HTTPSConnectionPool(host='our server', port=443): Read timed out. (read timeout=60)\"))"
  }
]
```

- this is not a user friendly message
- it leaks security sensitive info

Please replace with a message that clearly indicates that this is a temporary failure to retrieve logs, and user should just try again and report if persists. Or should log collection even work if this error occurs??

PR #318, opened 2023-01-23
#307 Better error message when log collection fails

JohanKJSchreurs:
This implements issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/307 .

### DIFF
--- a/openeogeotrellis/logs.py
+++ b/openeogeotrellis/logs.py
@@ -1,9 +1,10 @@
+import datetime as dt
 import json
 from typing import Iterable, Optional
 
 
 from elasticsearch import Elasticsearch, ConnectionTimeout
-from openeo.util import dict_no_none
+from openeo.util import dict_no_none, rfc3339
 from openeo_driver.errors import OpenEOApiException
 
 
@@ -12,18 +13,21 @@ ES_INDEX_PATTERN = "openeo-*-index-1m*"
 ES_TAGS = ["openeo"]
 
 
-def elasticsearch_logs(job_id: str, offset: Optional[str]) -> Iterable[dict]:
+def elasticsearch_logs(
+    job_id: str, create_time: dt.datetime, offset: Optional[str]
+) -> Iterable[dict]:
     try:
         search_after = None if offset in [None, ""] else json.loads(offset)
-        return _elasticsearch_logs(job_id, search_after)
+        return _elasticsearch_logs(job_id, create_time, search_after)
     except json.decoder.JSONDecodeError:
         raise OpenEOApiException(status_code=400, code="OffsetInvalid",
                                  message=f"The value passed for the query parameter 'offset' is invalid: {offset}")
 
 
-def _elasticsearch_logs(job_id: str, search_after: Optional[list]) -> Iterable[dict]:
+def _elasticsearch_logs(
+    job_id: str, create_time: dt.datetime, search_after: Optional[list]
+) -> Iterable[dict]:
     page_size = 100
-
     with Elasticsearch(ES_HOSTS) as es:
         while True:
             try:
@@ -34,6 +38,15 @@ def _elasticsearch_logs(job_id: str, search_after: Optional[list]) -> Iterable[d
                             "filter": [
                                 {"term": {"job_id": job_id}},
                                 {"terms": {"tags": ES_TAGS}},
+                                {
+                                    "range": {
+                                        "@timestamp": {
+                                            "format": "strict_date_optional_time",
+                                            "gte": rfc3339.datetime(create_time),
+                                            "lte": "now",
+                                        }
+                                    }
+                                },
                             ],
                         }
                     },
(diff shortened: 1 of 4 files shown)


openeogeotrellis/logs.py
```diff
@@ -34,6 +38,15 @@ def _elasticsearch_logs(job_id: str, search_after: Optional[list]) -> Iterable[d
                             "filter": [
                                 {"term": {"job_id": job_id}},
                                 {"terms": {"tags": ES_TAGS}},
+                                {
+                                    "range": {
+                                        "@timestamp": {
+                                            "format": "strict_date_optional_time",
```
bossie, line 65: `date_time` or `strict_date_time` might be a better fit (https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-date-format.html).

openeogeotrellis/logs.py
```diff
@@ -34,6 +38,15 @@ def _elasticsearch_logs(job_id: str, search_after: Optional[list]) -> Iterable[d
...
+                                            "gte": rfc3339.datetime(create_time),
+                                            "lte": "now",
```
bossie, line 67: I don't think the `"lte"` is needed.

JohanKJSchreurs: Indeed. I left it out now. It should work the same.

openeogeotrellis/logs.py
JohanKJSchreurs: OK, I changed it to strict_date_time.

FYI that code has moved a little bit, because I also made `create_time` an optional parameter of the function.
See: 
https://github.com/Open-EO/openeo-geopyspark-driver/blob/issue307_improve-error-message-when-log-collection-fails/openeogeotrellis/logs.py#L78-L96

jdries:
looks fine for me

openeogeotrellis/logs.py
```diff
@@ -1,67 +1,136 @@
...
+                {"terms": {"tags": ES_TAGS}},
+            ],
+        }
+    }
+    if create_time:
+        query["bool"]["filter"].append(
+            {
+                "range": {
+                    "@timestamp": {
+                        "format": "strict_date_optional_time",
```
bossie, line 91: Should be "strict_date_time".

openeogeotrellis/logs.py
```diff
@@ -1,67 +1,136 @@
...
+                    ],
+                    request_timeout=120,
+                )
+
+            except ConnectionTimeout as exc:
+                message = (
+                    "Temporary failure while retrieving logs (ConnectionTimeout). "
+                    + "Please try again and report this error if it persists."
+                )
+                raise OpenEOApiException(status_code=504, message=message)
```
bossie, line 119: We're losing valuable debugging information here i.e. the underlying ReadTimeoutError; is it possible to somehow retain this?

openeogeotrellis/logs.py
JohanKJSchreurs: Thanks for catching that one. I have mistakenly re-introduced the wrong format.

openeogeotrellis/logs.py
JohanKJSchreurs: Indeed it would be better to keep the info of the original exception somehow.

I will see if I can get and log that req_id from the view there, without too many complications.
I was also thinking we could log the entire ConnectionTimeout  as a warning. Maybe that is enough.
Generally, we have to be a bit careful with logging stuff inside code that processes logging, but to me it looks like that would be OK .

Additionally, would it be useful to use a `raise from` to link the original exception to the OpenEOApiException?
I suspect we should in fact avoid that, because I think it could still leak the security sensitive info that way.
While it may not be directly visible in the OpenEOApiException's message, the info is still present, and could be accessed.


Perhaps, something like this?

```python
            except ConnectionTimeout as exc:
                _logger.warning("Connection timeout while retrieving logs from Elasticsearch", exc_info=True)
                message = (
                    "Temporary failure while retrieving logs (ConnectionTimeout). "
                    + "Please try again and report this error if it persists."
                )
                raise OpenEOApiException(status_code=504, message=message) from exc
```

bossie: The idea was that we log the details but without a `job_id`; otherwise a user would simply be able to request the logs for this job ID again and still see the sensitive information.

I would include the request ID (`req_id`) because then we are able to take it from the user's error message and look up said log in Kibana.

> Generally, we have to be a bit careful with logging stuff inside code that processes logging, but to me it looks like that would be fine here.

In this case the logs that we are requesting (with a `job_id`) and the one that we write (without a `job_d`) will be mutually exclusive so I don't expect any issues.

Now that you mention it: maybe an easy fix without passing on the `req_id` all the way is to just chain the exceptions. If I understand this piece of code correctly, the chain of exceptions will already be logged with a `req_id` and the error message to the user will not include the `ConnectionTimeout` with the sensitive info:

https://github.com/Open-EO/openeo-python-driver/blob/9a1792ac98e33a37ecea681aedc00c2e3fef8eca/openeo_driver/views.py#L1236-L1240

openeogeotrellis/logs.py
JohanKJSchreurs: Done.

openeogeotrellis/logs.py
JohanKJSchreurs: Now I'm chaining the exception. 
Also added the req_id, it makes it easier to recognize when it is also in the message for the user.

There is a new test in `test_views.py, test_get_job_logs_with_connection_timeout`, which verifies a few important points about the logging, namely:
- the host URL is not leaked in the message for the user,
- the message format is the user friendly one we expect,
- the general log (the one not about the job) still contains the traceback information of ConnectionTimeouts.

https://github.com/Open-EO/openeo-geopyspark-driver/blob/65b1356011a8b58f70c510379f8682ea8fca324a/tests/test_views.py#L932-L1006

Let me know what you think.

bossie: Pushed some small changes myself, otherwise looked good.

- slightly changed error message to align it with other errors towards the client ("... (ref: r-b66a7c4f422440f6bf50db8b1d4fac0f)")
- more robust tests, in particular the last assert ("ConnectionTimeout" would also be present in the user's error message so a false negative)

bossie (approved):

Issue #325, opened 2023-01-27
refactor EtlApi for use in JobTracker v2


bossie:
Context: a batch job advertises credit usage that it gets from the ETL API (#272)

Key points:
-  currently only on Terrascope; this should work on CreoDIAS as well, but not in C-SCALE deployment
- `EtlApi` should fit within JobTracker v2 (#236)
- using `EtlApi` yes or no (see point 1) should depend on how JobTracker v2 is wired (think: an `on_job_done` listener)
- `EtlApi` seems very similar to `ElasticJobRegistry` (it authenticates with Keycloak and sends API requests); `ElasticJobRegistry` is already used in JobTracker v2 so that is convenient

bossie:
ETL API looks tailored towards YARN, i.e. the [state and status enums](https://etl.terrascope.be/docs/#/resources/ResourcesController_upsertResource), what about k8s?
> Op jouw eerste vraag - die state en status zijn niet zo super belangrijk. De status is enkel voor rapportering, maar daar doen we niet veel mee. De state zelf kan wel van belang zijn omdat we kunnen beslissen om al dat niet te credits af te trekken op basis van de state (bvb - killed en/of failed jobs wel of niet gaan aanrekenen)

Is the ETL API still relevant in the context of CDSE or is there another billing API?
> voorlopig zal de ETL API nog van toepassing zijn binnen CDSE. We hopen die er echter uit te krijgen zodat we rechtstreeks met de billing kunnen praten, maar voorlopig zou ik er nog vanuit gaan dat die gedeployed zal worden

Conclusion: a mapping of k8s state to YARN state should work.

bossie:
ETL API accepts these values for state = state: ACCEPTED, RUNNING, FINISHED, KILLED, FAILED, UNDEFINED
ETL API accepts these values for status = final_status: SUCCEEDED, FAILED, KILLED, UNDEFINED

:arrow_right:  map k8s states to ETL API states ACCEPTED, RUNNING, FINISHED, KILLED, FAILED, UNDEFINED

Also: https://github.com/Open-EO/openeo-geopyspark-driver/blob/98c0e8f95312957e07650be6a26425b1f917759c/openeogeotrellis/integrations/kubernetes.py

bossie:
Would like to address this first before proceeding: https://github.com/Open-EO/openeo-geopyspark-driver/pull/368

PR #331, opened 2023-02-02
325 refactor etlapi for use in jobtracker v2

bossie:
Implements https://github.com/Open-EO/openeo-geopyspark-driver/issues/325.


bossie:
Points 1-3 of https://github.com/Open-EO/openeo-geopyspark-driver/issues/325 are implemented by encapsulating the job costs calculation (in this case, the ETL API).

Do you think this might work @soxofaan ?

openeogeotrellis/job_tracker_v2.py
```diff
@@ -68,6 +80,72 @@ def get_job_metadata(self, job_id: str, user_id: str, app_id: str) -> _JobMetada
         raise NotImplementedError
 
 
+class JobCostsCalculator(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def calculate_costs(self, job_info: dict, job_metadata: _JobMetadata, result_metadata: dict) -> float:
+        raise NotImplementedError
+
+
+class EtlApiJobCostsCalculator(JobCostsCalculator):
```
soxofaan, line 89: If I understand correctly this in not only about calculating the cost, but also reporting/logging, right?

openeogeotrellis/job_tracker_v2.py
```diff
@@ -68,6 +80,72 @@ def get_job_metadata(self, job_id: str, user_id: str, app_id: str) -> _JobMetada
...
+    def __init__(self, etl_api: EtlApi, etl_api_access_token: str):
+        self._etl_api = etl_api
+        self._etl_api_access_token = etl_api_access_token
+
+    def calculate_costs(self, job_info: dict, job_metadata: _JobMetadata, result_metadata: dict) -> float:
```
soxofaan, line 94: It feels there is some conceptual redundancy between `job_info` and `job_metadata`. 
Is is easily possible to eliminate one in favor of the other?

openeogeotrellis/job_tracker_v2.py
soxofaan: Another thing: as far as I understand, the job cost calculation is quite separate from the job tracker itself. Could it make sense to move it  to its own module to have clearer separation of concerns?

soxofaan (commented):
overall this looks good I think

openeogeotrellis/job_tracker_v2.py
bossie: > If I understand correctly this in not only about calculating the cost, but also reporting/logging, right?

Yes: you give the ETL API some resource usage and it returns the costs in credits; then you patch those costs in the batch job entity.

> Another thing: as far as I understand, the job cost calculation is quite separate from the job tracker itself. Could it make sense to move it to its own module to have clearer separation of concerns?

Possibly. There's also the fact that a custom `JobCostsCalculator` implementation relies on the private `_JobMetadata` class so there's definitely room for improvement.

openeogeotrellis/job_tracker_v2.py
bossie: Also possibly. This was very ad-hoc to get things started.

openeogeotrellis/job_tracker_v2.py
bossie: I moved it to the `job_costs_calculator` module.

bossie:
Could take a quick look at this again @soxofaan ?

PR #347, opened 2023-02-20
Custom asset names

EmileSonneveld:
Implementing "filename_prefix" to prefix outputted asset names.

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -235,7 +235,7 @@ class TestDownload:
             "filename_prefix": "filenamePrefixTest"
         })
         assert 1 == len(res)
-        name, asset = res.popitem()
+        name, asset = next(iter(res.items()))
         assert Path(asset['href']).parent == tmp_path
         assert asset['nodata'] == -1
         assert asset['roles'] == ['data']
@@ -264,7 +264,7 @@ class TestDownload:
             "filename_prefix": "filenamePrefixTest",
         })
         assert len(res) == 3
-        name, asset = res.popitem()
+        name, asset = next(iter(res.items()))
         assert Path(asset['href']).parent == tmp_path
         assert asset['nodata'] == -1
         assert asset['roles'] == ['data']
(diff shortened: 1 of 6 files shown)


tests/test_download.py
```diff
@@ -195,78 +224,268 @@ def test_download_masked_json_reproject(self, tmp_path):
...
+                # "multidate": True,  # not used
+            }
+        )
+        # with open(self.test_write_assets_parameterize_path + test_name + ".json", 'w') as fp:
+        #     json.dump(assets, fp, indent=2)
+
+        name, asset = next(iter(assets.items()))
+        print("href of first asset: " + asset['href'])
+
+        # Special case for saveRDDTemporal as it only returns one case, but is designed to return multiple cases
```
EmileSonneveld, line 464: @jdries Here I was not sure why `saveRDDTemporal` only returns one file, while other code paths returned multiple files.
This is not something I changed, but only added coverage for. So, ok to just document it like this?

jdries: from what I can tell on sight, create_spacetime_layer creates a layer with only one date (self.now). Hence I would indeed inspect only one asset.
Test coverage would indeed be better if we have more than one date, can we change that method?

EmileSonneveld: That was indeed it. The hacky if-statement is removed now.

tests/test_download.py
```diff
@@ -89,15 +103,41 @@ def create_spacetime_layer(self):
 
         return TiledRasterLayer.from_numpy_rdd(LayerType.SPACETIME, rdd, metadata)
 
+    def create_spatial_layer(self):
```
jdries, line 106: could this be moved to datacube_fixtures.py, or even reuse an existing function from somewhere?

jdries (changes requested):
reviewed, only one minor comment

tests/test_download.py
EmileSonneveld: Done. If Ok, I can merge

Issue #292, opened 2022-12-14
log failed YARN container launch in Kibana


bossie:
Context: Jenkins launched a "cog" batch job but YARN was unable to start its container because it was unable to pull the Docker image from Artifactory.

The JobTracker was able to determine that the YARN job FAILED and marked the OpenEO batch job as "error" and reported this in the logs (good); the reason why it failed, however, was not in the logs because our logging is obviously not set up correctly if the application fails to run in the first place.

![failed_container_launch](https://user-images.githubusercontent.com/1032518/207575010-f87729fa-503d-4f15-b8b3-ad379c93b7e4.png)

bossie:
Seeing as

- the JobTracker already calls `yarn application -status` to determine the outcome of the batch job
- that command's output also contains the error that we're looking for

it should be pretty easy to make JobTracker log that error as well.

```
[vdboschj@vdboschj ~]$ yarn application -status application_1670152552564_21143
22/12/14 11:33:20 INFO client.RequestHedgingRMFailoverProxyProvider: Created wrapped proxy for [rm1, rm2]
22/12/14 11:33:20 INFO client.AHSProxy: Connecting to Application History server at epod-master3.vgt.vito.be/192.168.207.58:10200
22/12/14 11:33:20 INFO client.RequestHedgingRMFailoverProxyProvider: Looking for the active RM in [rm1, rm2]...
22/12/14 11:33:21 INFO client.RequestHedgingRMFailoverProxyProvider: Found active RM [rm2]
22/12/14 11:33:21 INFO conf.Configuration: resource-types.xml not found
22/12/14 11:33:21 INFO resource.ResourceUtils: Unable to find 'resource-types.xml'.
Application Report : 
	Application-Id : application_1670152552564_21143
	Application-Name : openEO batch_cog_j-7b184aa29465478ebf1764c143ed5c82_user jenkins
	Application-Type : SPARK
	User : jenkins
	Queue : lowlatency
	Application Priority : 0
	Start-Time : 1671010049075
	Finish-Time : 1671010311173
	Progress : 0%
	State : FAILED
	Final-State : FAILED
	Tracking-URL : https://epod-master2.vgt.vito.be:8090/cluster/app/application_1670152552564_21143
	RPC Port : -1
	AM Host : N/A
	Aggregate Resource Allocation : 536592 MB-seconds, 524 vcore-seconds
	Aggregate Resource Preempted : 0 MB-seconds, 0 vcore-seconds
[...]

PR #348, opened 2023-02-21
Issue292 log failed yarn container launch in kibana

JohanKJSchreurs:
Implements issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/292


Jenkinsfile
```diff
@@ -23,7 +23,7 @@ pythonPipeline {
   extras_require = 'dev'
   upload_dev_wheels = false
   pep440 = true
-  venv_rpm_deps = ['gcc-c++', 'kstart']
+  venv_rpm_deps = ['gcc-c++', 'kstart', 'krb5-devel']
```
JohanKJSchreurs, line 26: FYI:
I added the krb5-devel package, which provides `krb5-config`, because requests_gssapi needs it during its pip installation. I noticed that when the Jenkins build failed.
(requests_gssapi is used for Kerberos authentication)

openeogeotrellis/configparams.py
```diff
@@ -28,6 +28,13 @@ def __init__(self, env=os.environ):
             "epod-master1.vgt.vito.be:6668,epod-master2.vgt.vito.be:6668,epod-master3.vgt.vito.be:6668",
         )
 
+        self.yarn_rest_api_base_url = env.get(
+            "YARN_REST_API_BASE_URL", "https://epod-master1.vgt.vito.be:8090"
+        )
+        self.yarn_auth_use_kerberos = smart_bool(
```
bossie, line 34: Do we need this config param? AFAIK we always require Kerberos authentication to access YARN.

openeogeotrellis/configparams.py
```diff
@@ -28,6 +28,13 @@ def __init__(self, env=os.environ):
...
+            env.get("YARN_AUTH_USE_KERBEROS", "yes")
+        )
```
bossie, line 36: Do we need this `yarn_auth_use_kerberos` config param? AFAIK we always require Kerberos authentication to access YARN.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
 class YarnStatusGetter(JobMetadataGetterInterface):
     """YARN app status getter"""
 
+    def get_application_url(self, application_id: str) -> str:
+        return url_join(
+            ConfigParams().yarn_rest_api_base_url,
+            f"/ws/v1/cluster/apps/{application_id}",
+        )
```
bossie, line 94: Instead of grabbing the URL from the environment in this method, I would define a constructor for `YarnStatusGetter` that takes this URL and stores it in a member variable. This makes `YarnStatusGetter`'s dependency on that URL explicit.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -16,6 +16,8 @@
 from typing import Callable, List, NamedTuple, Optional, Union
```
bossie, line 16: The new implementation rendered some imports unused so they can be removed.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+            f"/ws/v1/cluster/apps/{application_id}",
+        )
+
+    def get_authentication_provider(self):
+        auth = None
+        if ConfigParams().yarn_auth_use_kerberos:
+            # TODO: HTTPKerberosAuth is deprecated, this is a utility class for compatibility with requests_kerberos.
+            #   Find the better way to do this, using requests_gssapi directly.
+            # Have to set mutual_authentication to REQUIRED or it just skips kerberos auth it seems.
+            auth = requests_gssapi.HTTPKerberosAuth(
```
bossie, line 102: Is it possible to also address this TODO? `HTTPSPNEGOAuth` seems to be the replacement.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+                mutual_authentication=requests_gssapi.REQUIRED
+            )
+        return auth
```
bossie, line 105: If `yarn_auth_use_kerberos` is not necessary (see above) we can remove this method and simply initialize `auth`.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
-            stdout = e.stdout.decode()
-            if "doesn't exist in RM or Timeline Server" in stdout:
-                raise AppNotFound(stdout) from None
-            else:
-                raise
-        return self.parse_application_report(application_report)
+        # TODO: Are there any cases were we *do* need to provide credentials to the Kerberos authentication?
+        #   At present we are assuming there is already a Kerberos ticket because kinit was run before
+        #   the geopyspark driver process was started.
+        #   AFAIK that means we don't have to provide user & pw which is in fact the whole point.
```
bossie, line 111: Indeed, the `kinit` takes care of this so this TODO can be removed.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+        #         "javaClassName": "org.apache.hadoop.yarn.webapp.NotFoundException"
+        #     }
+        # }
+        if response.status_code == 404:
+            remote_exc = response.json().get("RemoteException", {})
+            message = remote_exc.get(
+                "message",
+                "Unknown error. Error response did not contain the message property.",
+            )
+            raise AppNotFound(message) from None
```
bossie, line 134: Can we just pass the entire response as text to `AppNotFound`? We already established that the application does not exist (404) and this seems to just remove context for no clear reason.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+        """Parse millisecond timestamp from app report and return as rfc3339 date (or None)"""
+        if epoch_millis == 0:
+            return None
+        utc_datetime = dt.datetime.utcfromtimestamp(epoch_millis / 1000)
+        return rfc3339.datetime(utc_datetime)
+
+    @classmethod
+    def parse_application_response(cls, json: dict) -> _JobMetadata:
+        if not json:
+            raise YarnAppReportParseException("Response is empty")
```
bossie, line 151: Unnecessary check because `response.json()` will already have failed.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+
+        # Handle a corrupt response that comes through as a string, not valid JSON.
+        if isinstance(json, str):
+            raise YarnAppReportParseException(
+                f"Response is corrupt, not valid JSON. Received following string: {json!r}"
+            )
```
bossie, line 157: I would just assume that the response is a JSON object that will get parsed to a dict. `response.json()` will have raised an exception for any unexpected response anyway.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
 
-    def parse_application_report(self, report: str) -> _JobMetadata:
-        try:
-            props = dict(re.findall(r"^\t(.+?) : (.+)$", report, flags=re.MULTILINE))
+        report = json.get("app", {})
+        required_keys = ["state", "finalStatus", "startedTime", "finishedTime"]
+        missing_keys = [k for k in required_keys if k not in report]
+        if missing_keys:
+            raise YarnAppReportParseException(
+                f"JSON response is missing following required keys: {missing_keys}"
```
bossie, line 164: I would include the JSON response itself as well, not just the missing keys.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
-                state=props["State"], final_state=props["Final-State"]
+                state=report["state"], final_state=report["finalStatus"]
             )
+            # Log the diagnostics if there was an error.
+            # In particular, sometimes YARN can't launch the container and then the
+            # field 'diagnostics' provides more info why this failed.
+            diagnostics = report.get("diagnostics", None)
+            if job_status == JOB_STATUS.ERROR and diagnostics:
+                _log.error(
+                    f"YARN application status reports error diagnostics: diagnostics: {diagnostics}"
```
bossie, line 177: "diagnostics: " is duplicated.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +87,112 @@ class YarnAppReportParseException(Exception):
...
+                usage = {}
+                memory_seconds = report.get("memorySeconds")
+                vcore_seconds = report.get("vcoreSeconds")
+                if memory_seconds is not None:
+                    usage["memory"] = {
+                        "value": int(memory_seconds),
+                        "unit": "mb-seconds",
+                    }
+                if vcore_seconds is not None:
+                    usage["cpu"] = {"value": int(vcore_seconds), "unit": "cpu-seconds"}
```
bossie, line 195: Not exactly the behavior that we had before, but I would just try to log memory seconds and vcore seconds if they're available, regardless of whether or not one is missing.

tests/conftest.py
```diff
@@ -38,6 +38,10 @@ def pytest_configure(config):
     os.environ["OPENSEARCH_ENRICH"] = "no"
     os.environ["ASYNC_TASKS_KAFKA_BOOTSTRAP_SERVERS"] = "kafka01.test:6668"
 
+    # For testing job_tracker_v2: TestYarnStatusGetter and TestYarnJobTracker
+    os.environ["YARN_REST_API_BASE_URL"] = "https://openeo.test"
+    os.environ["YARN_AUTH_USE_KERBEROS"] = "no"
```
bossie, line 43: Setting `YARN_REST_API_BASE_URL` is unnecessary if you pass the URL to the constructor (see above).
Setting `YARN_AUTH_USE_KERBEROS` is unnecessary if you're not doing any actual HTTP requests.

openeogeotrellis/configparams.py
JohanKJSchreurs: Good question.
The actual purpose here was to be able to turn Kerberos authentication off during unit tests. However that does not really mean that we need to be able to configure this for a production setup.

I will try to find a better solution. That may indeed simplify a few things, as your other comments below also indicated.

bossie, line 36: `YARN_AUTH_USE_KERBEROS` is basically to skip Kerberos auth in unit tests, right?

tests/test_job_tracker_v2.py
```diff
@@ -751,9 +927,65 @@ def test_parse_application_report_running(self):
...
+    def test_parse_application_response_empty(self):
         with pytest.raises(YarnAppReportParseException):
-            _ = YarnStatusGetter().parse_application_report(report="")
+            _ = YarnStatusGetter.parse_application_response(json={})
+
+    def test_get_authentication_provider_use_kerberos_on(self, monkeypatch):
+        """Simple check that the config has the desired effect: we get the expected auth object."""
+        monkeypatch.setenv("YARN_AUTH_USE_KERBEROS", "yes")
+        auth = YarnStatusGetter().get_authentication_provider()
+        assert isinstance(auth, requests_gssapi.HTTPKerberosAuth)
```
bossie, line 938: If you pass auth to the constructor of `YarnStatusGetter` (be it a `bool`/`HTTPKerberosAuth`/`None`) this test is covered by `test_kerberos_auth_should_fail_when_there_is_no_kerberos_setup` below so you can get rid of this test and the `get_authentication_provider` method as well.

tests/test_job_tracker_v2.py
```diff
@@ -751,9 +927,65 @@ def test_parse_application_report_running(self):
...
+
+    def test_get_authentication_provider_use_kerberos_off(self, monkeypatch):
+        """Simple check that the config has the desired effect: authentication should be None."""
+        monkeypatch.setenv("YARN_AUTH_USE_KERBEROS", "no")
+        assert YarnStatusGetter().get_authentication_provider() is None
```
bossie, line 943: Same remark: passing `False`/`None` to the constructor makes it so that you can get rid of method `get_authentication_provider` and this test as well because it's probably already covered elsewhere.

tests/test_job_tracker_v2.py
```diff
@@ -751,9 +927,65 @@ def test_parse_application_report_running(self):
...
+
+    def test_kerberos_auth_should_fail_when_there_is_no_kerberos_setup(
```
bossie, line 945: I don't think this is a good name because this is not the behavior that you're trying to assert, it's just the way it's implemented.

Maybe `test_kerberos_auth_is_actually_attempted()` is better (like the docstring describes).

bossie (changes requested):
Added remarks inline.

openeogeotrellis/configparams.py
JohanKJSchreurs: Yes, exactly.

bossie:
One more thing that has to be verified in the context of replacing the YARN CLI with `requests`: the YARN resource manager runs in an HA setup so you have two nodes: https://epod-master1.vgt.vito.be:8090 and https://epod-master2.vgt.vito.be:8090, where one is active and the other one is passive.

If you send a request to epod-master1.vgt.vito.be and it is the active instance, it will Just Work. If, on the other hand, epod-master1.vgt.vito.be is currently passive, it will respond with a redirect to active node epod-master2.vgt.vito.be and you have to make sure that `requests` handles this redirect.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: That's better indeed.
Done.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Done: now its using requests_gssapi.HTTPSPNEGOAuth directly, and only where we really need to use authentication.
(So not in the unit tests)

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Now the constructor accepts an optional auth object, which is None by default.
That has helped to simplify quite a bit.

- Basically, the config param for yarn_auth_use_kerberos and related code has been removed entirely. 
- Unit tests leave just out the auth param
- And only the CLI app passes the Kerberos auth to the YarnStatusGetter constructor here:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/3920e20dda0b1ef027148d0146658b696c4c87a9/openeogeotrellis/job_tracker_v2.py#L537-L543

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Indeed. Removed it.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Done.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Done.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Good idea, much simpler code. 
Done.

tests/conftest.py
JohanKJSchreurs: I was indeed able to eliminate them by following your other suggestions.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Removed the unused imports.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Done.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Done.

openeogeotrellis/configparams.py
JohanKJSchreurs: Meanwhile I removed the environment variable YARN_AUTH_USE_KERBEROS and related code.

tests/test_job_tracker_v2.py
JohanKJSchreurs: Indeed. In the end I was able to remove all three of these tests.

tests/test_job_tracker_v2.py
JohanKJSchreurs: Done.

tests/test_job_tracker_v2.py
JohanKJSchreurs: I ended up removing this test too.
After the refactoring it was not a very useful test anymore because there was only one place left where we need to instantiate the Kerberos authentication.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: This was bothering me a bit too.

But while trying to eliminate this check, I realized that a single string is apparently also valid JSON and because of that  response.json() doesn't raise an exception.

Also there was an existing test case for a corrupted response that just some piece of text. Though that was originally a test case for corrupted output of the yarn CLI command, and perhaps that test case is not truly useful anymore.

So I took a view and clarified that the expected type for the parameter `json` can be dict or str, and made the error message a bit more clear.

bossie: I'm not sure how response.json() behaves but I should hope that it raises an exception on strings not enclosed in double quotes.

JohanKJSchreurs: As we discussed this morning, I will move this check one layer up to the calling method.
That is a more appropriate place to validate this, should we ever get incorrect responses.


I also dug a little bit deeper into this because I wanted to know why a response body that contained - _seemingly_ - invalid JSON was slipping through the cracks. 

### In short

The problem was in the mock, not in requests.json()

Initially, I was surprised when I saw that requests.Response.json() did not raise an exception when the response was seemingly a string without double quotes around it. But the unit test was not set up correctly to test that.

The mock produced different data than I thought it would, and it actually gave requests a string ***inside double quotes***. So requests received valid JSON and there is no lurking problem with requests.json() after all.

What is more, requests.json() uses json.loads(), so it does behave the same as json.loads(), as we would expect.


For some reason the mock always adds double quotes around the response text if the response body is just a string, probably because we use in a way that is meant to produce JSON responses specifically.

If we wanted to simulate this particular (edge) case, where requests should detect invalid JSON, I think we would need to use requests_mock in a different way and create a raw text response. But that is a bit more cumbersome, and I think I will leave it at that.

openeogeotrellis/job_tracker_v2.py
```diff
@@ -8,14 +8,13 @@
...
-import re
-import subprocess
 from decimal import Decimal
 from pathlib import Path
-from subprocess import CalledProcessError
-from typing import Callable, List, NamedTuple, Optional, Union
+from typing import Any, List, NamedTuple, Optional, Union
 
 import requests
+import requests_gssapi
```
soxofaan, line 16: I think we should make this an optional dependency under a "yarn" extra, 
because we don't need this on kube deploys

setup.py
```diff
@@ -85,6 +85,7 @@
         'boto3==1.16.25',
         'hvac==1.0.2',
         "pyarrow>=1.0.0",  # For pyspark.pandas
+        "requests-gssapi>=1.2.3",  # For Kerberos authentication
```
soxofaan, line 88: as noted above: I think we should make this an optional dependency under a "yarn" extra, 
because we don't need this on kube deploys

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: I made the dependency optional as an extra requirement for yarn.

And here in job_tracker_v2 module I kept the import at the top of the module rather than doing a local import in the function that uses it. I think that makes it easier to keep a clear overview of what the module needs, and what it optionally uses.

Therefore it sets requests_gssapi to None if it is not available.

https://github.com/Open-EO/openeo-geopyspark-driver/blob/a8c943ada76f5ab4a414e6b7ec30570bf1e77295/openeogeotrellis/job_tracker_v2.py#L17-L21

openeogeotrellis/job_tracker_v2.py
```diff
@@ -85,48 +84,105 @@ class YarnAppReportParseException(Exception):
...
+    @staticmethod
+    def _ms_epoch_to_date(epoch_millis: int) -> Union[str, None]:
+        """Parse millisecond timestamp from app report and return as rfc3339 date (or None)"""
+        if epoch_millis == 0:
+            return None
+        utc_datetime = dt.datetime.utcfromtimestamp(epoch_millis / 1000)
+        return rfc3339.datetime(utc_datetime)
+
+    @classmethod
+    def parse_application_response(cls, json: dict) -> _JobMetadata:
```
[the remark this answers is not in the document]
JohanKJSchreurs: Done.

setup.py
JohanKJSchreurs: Done.

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: > One more thing that has to be verified in the context of replacing the YARN CLI with `requests`: the YARN resource manager runs in an HA setup so you have two nodes: https://epod-master1.vgt.vito.be:8090 and https://epod-master2.vgt.vito.be:8090, where one is active and the other one is passive.
> 
> If you send a request to epod-master1.vgt.vito.be and it is the active instance, it will Just Work. If, on the other hand, epod-master1.vgt.vito.be is currently passive, it will respond with a redirect to active node epod-master2.vgt.vito.be and you have to make sure that `requests` handles this redirect.

The requests library should already handle this. 

In the API docs we find that `requests.get(...) requests.post(...)` etc have the parameter  
`allow_redirects`  and that it defaults to True

See also: https://requests.readthedocs.io/en/latest/api/

JohanKJSchreurs:
> One more thing that has to be verified in the context of replacing the YARN CLI with `requests`: the YARN resource manager runs in an HA setup so you have two nodes: https://epod-master1.vgt.vito.be:8090 and https://epod-master2.vgt.vito.be:8090, where one is active and the other one is passive.
> 
> If you send a request to epod-master1.vgt.vito.be and it is the active instance, it will Just Work. If, on the other hand, epod-master1.vgt.vito.be is currently passive, it will respond with a redirect to active node epod-master2.vgt.vito.be and you have to make sure that `requests` handles this redirect.

The requests library should already handle this.

In the API docs we find that requests.get(...) requests.post(...) etc have the parameter
allow_redirects and that it defaults to True

See also: https://requests.readthedocs.io/en/latest/api/

setup.py
soxofaan: FYI: in 1800b1e17ea0ca5821f7bc418ade19c71b33bc4f I added an empty "yarn" extra already on master so that we can already start including it in the deploy pipelines. Otherwise we risk a window of tricky deploy steps.

It will probably introduce a trivial merge conflict in this PR however

openeogeotrellis/job_tracker_v2.py
JohanKJSchreurs: Turns out it wasn't that complicated to improve the unit test after all.

So now we cover both cases: 
- Invalid JSON gets caught by responses.json() and raises a JSONDecodeError as expected.
- If the response contains valid JSON that is not a dict, that will be caught by our own code and it raises YarnAppReportParseException

bossie (approved):

PR #349, opened 2023-02-23
Issue Open-EO/openeo-geotrellis-extensions#72 Added projection extens…

JohanKJSchreurs:
# Add projection extension metadata to batch job results

Implements https://github.com/Open-EO/openeo-geotrellis-extensions/issues/72

## Dependencies

**Updated:** The remark below about changes in the dependencies no longer applies because these changes have been reverted: 

~~Notable changes in dependencies:~~
- ~~added  python package `gdal==3.4.*`~~
- ~~and for Jenkins that means it also needs gdal-devel  (v3.4) because the python installation apparently needs the gdal-config command.~~

**N.B.**
That still means there are commits on `setup.py` and `jenkinsfile` that modified and later reverted those modification on these two files.

Namely, I identified the following commits:

Added pypi gdal to setup.py 
https://github.com/Open-EO/openeo-geopyspark-driver/pull/349/commits/34bc81e7d46723f41a8f204e0b04ff4cce48d154

Added gdal-devel to jenkinsfile:
https://github.com/Open-EO/openeo-geopyspark-driver/pull/349/commits/45be667cb857642d65a2288da910992bb13eb46b

Merge commit resolving incoming addition from master and gdal-devel from the feature branch in jenkinsfile
https://github.com/Open-EO/openeo-geopyspark-driver/pull/349/commits/6ecdf28f679ff95970de9aee4ae886d634db7beb

Removed both gdal from setup.py and gdal-devel from jenkinsfile
https://github.com/Open-EO/openeo-geopyspark-driver/pull/349/commits/bcbb2e8ad99a253240160547ef3589b3f4a0689e


openeogeotrellis/deploy/batch_job.py
```diff
@@ -211,6 +270,238 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+        # TODO: check if this way to combine the corners also handles 3D bounding boxes correctly.
+        #   Need a correct example to test with.
+        lole = corner_coords["lowerLeft"]
+        upri = corner_coords["upperRight"]
+        proj_metadata["proj:bbox"] = [*lole, *upri]
+
+    return proj_metadata
+
+
+def extract_crs_epsg_code_from_wkt_string(crs_as_wkt: str) -> str:
```
jdries, line 409: I would use CRS.from_wkt or CRS.from_user_input and then get epsg code out of it (CRS.to_epsg)

jdries (changes requested):
One remark about epsg lookup, looks good to me otherwise.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: Done.

Switched to using the CRS method from `pyproj.CRS` and removed `extract_crs_epsg_code_from_wkt_string`. 
This should indeed be more robust.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -191,8 +194,65 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+                bboxs = {tuple(m["proj:bbox"]) for m in projection_metadata.values()}
+                same_bbox_all_assets = len(bboxs) == 1
+
+                shapes = {tuple(m["proj:shape"]) for m in projection_metadata.values()}
+                same_shapes_all_assets = len(shapes) == 1
+
+            if same_epsg_all_assets and same_bbox_all_assets and same_shapes_all_assets:
+                # TODO: Should we overwrite existing values for epsg and bbox, or keep what is already there?
+                if not epsg and not metadata.get("epsg"):
+                    epsg = epsgs.pop()
```
bossie, line 247: I'm getting a warning on this line:
> Local variable 'epsgs' might be referenced before assignment

openeogeotrellis/deploy/batch_job.py
```diff
@@ -191,8 +194,65 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+                if not metadata.get("bbox"):
+                    metadata["bbox"] = list(bboxs.pop())
```
bossie, line 249: Same warning for `bboxs`.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -211,6 +271,208 @@ def epsg_code(gps_crs) -> Optional[int]:
 
     logger.info("wrote metadata to %s" % metadata_file)
 
+
+def read_projection_extension_metadata(asset_path: str) -> dict:
```
bossie, line 275: Return type should be an `Optional[dict]`.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: I have cleaned up the logic so this could be avoided.
There was an if statement that could be eliminated, and simplifying this made the intention a bit more clear as well.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: Done, same as for epsgs.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -191,8 +194,65 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+                shapes = {tuple(m["proj:shape"]) for m in projection_metadata.values()}
+                same_shapes_all_assets = len(shapes) == 1
+
+            if same_epsg_all_assets and same_bbox_all_assets and same_shapes_all_assets:
+                # TODO: Should we overwrite existing values for epsg and bbox, or keep what is already there?
+                if not epsg and not metadata.get("epsg"):
+                    epsg = epsgs.pop()
+                if not metadata.get("bbox"):
+                    metadata["bbox"] = list(bboxs.pop())
+                metadata["proj:shape"] = list(shapes.pop())
```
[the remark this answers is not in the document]
JohanKJSchreurs: Done, same as for epsgs.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: Indeed.

In fact I dug a bit deeper with mypy and mypy found some more potential issues, so I decided to make the return and argument types more consistent overal.

I got rid of returning None because we can just as well return an empty dict in these cases. The check is the same but we don't do anything special with None. 
That avoids extra checks for None inside the functions, and it avoids that function arguments also need to have type Optional[dict] .

Further, I introduced two type aliases because there were a lot of Dict[str, Any] everywhere and that made it confusing what was the difference between dicts at the input (coming from gdal.Info) versus dicts at the output, namely the projection metadata in STAC format.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -211,6 +271,208 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+    """
+    # Make gdal raise exception, otherwise it just writes errors on stdout.
+
+    gdal_info = read_gdal_info(asset_path)
+    if not gdal_info:
+        return None
+    return parse_projection_extension_metadata(gdal_info)
+
+
+def read_gdal_info(asset_uri: str) -> dict:
```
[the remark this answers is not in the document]
JohanKJSchreurs: Fixed in the same way as the other problems with incorrect return type.  
( Comment above: https://github.com/Open-EO/openeo-geopyspark-driver/pull/349#discussion_r1145854491 )

openeogeotrellis/deploy/batch_job.py
```diff
@@ -211,6 +271,208 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+        # TODO: check if this way to combine the corners also handles 3D bounding boxes correctly.
+        #   Need a correct example to test with.
+        lole = corner_coords["lowerLeft"]
+        upri = corner_coords["upperRight"]
+        proj_metadata["proj:bbox"] = [*lole, *upri]
+
+    return proj_metadata
+
+
+def _process_gdalinfo_for_netcdf_subdatasets(gdal_info: dict) -> dict:
```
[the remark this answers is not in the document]
JohanKJSchreurs: Fixed in the same way as the other problems with incorrect return type.  
( Comment above: https://github.com/Open-EO/openeo-geopyspark-driver/pull/349#discussion_r1145854491 )

openeogeotrellis/deploy/batch_job.py
```diff
@@ -211,6 +271,208 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+    # will not be processed.
+    if gdal_info.get("driverShortName") != "netCDF":
+        return None
+    if "SUBDATASETS" not in gdal_info.get("metadata", {}):
+        return None
+
+    sub_datasets = {}
+    for key, sub_ds_uri in gdal_info["metadata"]["SUBDATASETS"].items():
+        if key.endswith("_NAME"):
+            sub_ds_gdal_info = gdal_info = read_gdal_info(sub_ds_uri)
```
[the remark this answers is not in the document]
JohanKJSchreurs: That was a big typo that I missed, because gdal_info should not have been there.
Have removed it.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -191,8 +194,65 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+            # When the projection metadata is the same for all assets, then set it at the item level only.
+            # This makes it easier to read, so we see at a glance that all bands have the same proj metadata.
+            projection_metadata = {}
+
+            # We also check whether any asset file was missing or its projection metadata could not be read.
+            # In that case we never write the projection metadata at the item level.
+            missing_files_or_projection_md = False
+
+            for asset_path in asset_metadata.keys():
+                if not Path(asset_path).exists():
```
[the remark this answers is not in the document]
JohanKJSchreurs: Removed this check.

( See also reply to your other comment below: https://github.com/Open-EO/openeo-geopyspark-driver/pull/349#discussion_r1145911067 )

openeogeotrellis/deploy/batch_job.py
```diff
@@ -191,8 +194,65 @@ def epsg_code(gps_crs) -> Optional[int]:
...
+                    # Path should exist so log an error.
+                    # We should only have missing files during unit tests that don't actually use files.
+                    # However, if this happens in production we should see it in the logs.
```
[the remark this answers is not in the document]
JohanKJSchreurs: Correction, I spoke to soon:
13 tests were in fact failing without the check for file existence so that means only 7 tests would be affected, not 20.
This was a test on my system, not the CI.

At first sight, some of these 13 tests rely on data that has to be mounted to the file system, and I currently don't have that on Windows Subsystem for Linux. Hence, file not found would be expected in that case.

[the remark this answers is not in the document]
JohanKJSchreurs: I have tried to check if a batch job could reach the point of extracting metadata if the asset file does not exist.

It looks like in practices the job would fail before that point, because `add_permissions(...)` would be called on the asset files, `run_job`

https://github.com/Open-EO/openeo-geopyspark-driver/blob/3d901f75a28d07ae3e03dfd127c2bcc862cf7af3/openeogeotrellis/deploy/batch_job.py#L713-L742

So, I think for now it is enough to log a warning when gdal.Info can not get metadata.

If we do add a check to verify the asset file exists, then we will have to fix a few unit tests and in future tests we would have to create some more fake files just so the thing won't raise an exception.

But if we do want that more strict checking, then let's create a follow up issue for that.

Issue #321, opened 2023-01-25
extent in STAC result metadata should be lat lon


jdries:
For a job (j-f4a4a6605c7c4b8d8ec8e79a2dbd408e) that specifies the bounding box in EPSG:3035, I get the metadata below in job results. I believe stac extents are always in EPSG:4326. The extent in asset crs can go in proj:bbox (but the filter_bbox crs is not necessarily asset crs...)

```
"extent": {
    "spatial": {
      "bbox": [
        [
          2780000.0,
          1820000.0,
          2800000.0,
          1840000.0
        ]
      ]
    },
```

This line of code seems to ignore crs entirely:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/d7aa3755f3bd5dfbd4d19db9de20d66586b66648/openeogeotrellis/deploy/batch_job.py#L100

That would need to be changed...

PR #360, opened 2023-03-02
Issue321 extent in stac result metadata should be lat lon

JohanKJSchreurs:
Implements issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/321

### DIFF
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -97,10 +97,6 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
     extents = [sc["spatial_extent"] for _, sc in source_constraints if "spatial_extent" in sc]
     if(len(extents) > 0):
         spatial_extent = spatial_extent_union(*extents)
-        if spatial_extent["crs"] != "EPSG:4326":
-            spatial_extent = reproject_bounding_box(
-                spatial_extent, from_crs=None, to_crs="EPSG:4326"
-            )
         bbox = [spatial_extent[b] for b in ["west", "south", "east", "north"]]
         if all(b is not None for b in bbox):
             polygon = Polygon.from_bounds(*bbox)
@@ -141,6 +137,14 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
 
     links = tracer.get_metadata_links()
     links = [link for k, v in links.items() for link in v]
+
+    if bbox:
+        # Reproject the spatial extent if it is not in lat-lon / WGS84.
+        if spatial_extent["crs"] != "EPSG:4326":
+            spatial_extent = reproject_bounding_box(
+                spatial_extent, from_crs=None, to_crs="EPSG:4326"
+            )
+            bbox = [spatial_extent[b] for b in ["west", "south", "east", "north"]]
 
     # TODO: dedicated type?
     # TODO: match STAC format?
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -151,6 +151,7 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
     return {
         'geometry': geometry,
         'bbox': bbox,
+        # 'proj:bbox': proj_bbox,
         'area': {'value': area, 'unit': 'square meter'} if area else None,
         'start_datetime': start_date,
         'end_datetime': end_date,
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -23,7 +23,7 @@ from openeo_driver.dry_run import DryRunDataTracer
 from openeo_driver.save_result import ImageCollectionResult, JSONResult, MultipleFilesResult, SaveResult, NullResult, \
     MlModelResult
 from openeo_driver.users import User
-from openeo_driver.util.geometry import spatial_extent_union
+from openeo_driver.util.geometry import spatial_extent_union, reproject_bounding_box
 from openeo_driver.util.logging import BatchJobLoggingFilter, user_id_trim, get_logging_config, setup_logging, \
     LOGGING_CONTEXT_BATCH_JOB
 from openeo_driver.util.utm import area_in_square_meters
@@ -97,6 +97,10 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
     extents = [sc["spatial_extent"] for _, sc in source_constraints if "spatial_extent" in sc]
     if(len(extents) > 0):
         spatial_extent = spatial_extent_union(*extents)
+        if spatial_extent["crs"] != "EPSG:4326":
+            spatial_extent = reproject_bounding_box(
+                spatial_extent, from_crs=None, to_crs="EPSG:4326"
+            )
         bbox = [spatial_extent[b] for b in ["west", "south", "east", "north"]]
         if all(b is not None for b in bbox):
             polygon = Polygon.from_bounds(*bbox)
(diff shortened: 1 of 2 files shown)


soxofaan (commented):
just minor note, but looks good to me

openeogeotrellis/deploy/batch_job.py
```diff
@@ -138,6 +138,14 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
     links = tracer.get_metadata_links()
     links = [link for k, v in links.items() for link in v]
 
+    if bbox:
+        # Reproject the spatial extent if it is not in lat-lon / WGS84.
+        if spatial_extent["crs"] != "EPSG:4326":
```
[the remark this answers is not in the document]
JohanKJSchreurs: Done.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -120,6 +125,8 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
             logger.warning("Multiple aggregate_spatial geometries: {c}".format(c=len(aggregate_spatial_geometries)))
         agg_geometry = aggregate_spatial_geometries[0]
         if isinstance(agg_geometry, BaseGeometry):
+            # We only allow EPSG:4326 for the BaseGeometry case.
+            bbox_crs = "EPSG:4326"
             bbox = agg_geometry.bounds
             geometry = mapping(agg_geometry)
             area = area_in_square_meters(agg_geometry, "EPSG:4326")
```
bossie, line 132: Would replace `"EPSG:4326"` with `bbox_crs`.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -128,16 +135,47 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
...
+            if aggregate_spatial_geometries:
+                new_spatial_extent = {
+                    "west": bbox[0],
+                    "south": bbox[1],
+                    "east": bbox[2],
+                    "north": bbox[3],
+                    "crs": bbox_crs,
+                }
+            else:
+                new_spatial_extent = spatial_extent
```
bossie, line 173: I'm getting a warning that `spatial_extent` might not be initialized; you might want to account for the case where there is no extent.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -128,16 +135,47 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
...
+    if bbox:
+        # Note that if the bbox comes from the aggregate_spatial_geometries, then we may
+        # get a pyproy CRS object instead of an EPSG code.That's fine. In that case we
+        # can simply go ahead with the reprojection.
+        # If the CRS happens to be EPSG:4326 after all then it will be a no-operation.
+        # It is difficult to list and check all possible variants of pyproj CRS objects
+        # that are actually all the exact same EPSG:4326 CRS, and it isn't really necessary.
+        if bbox_crs not in [4326, "EPSG:4326", "epsg:4326"]:
+            if aggregate_spatial_geometries:
+                new_spatial_extent = {
```
bossie, line 165: Could you give this a more descriptive name?

openeogeotrellis/deploy/batch_job.py
```diff
@@ -120,6 +125,8 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
             logger.warning("Multiple aggregate_spatial geometries: {c}".format(c=len(aggregate_spatial_geometries)))
         agg_geometry = aggregate_spatial_geometries[0]
         if isinstance(agg_geometry, BaseGeometry):
+            # We only allow EPSG:4326 for the BaseGeometry case.
```
bossie, line 128: Is this because the DryRunDataTracer will always return BaseGeometries in EPSG:4326?

bossie (changes requested):
Added remarks inline.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: I missed that one. 
Done.

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: No, this was more of a pragmatic decision to keep things simple and prevent complicated cases with different CRSs.

The BaseGeometries are objects from Shapely, but Shapely doesn't really support CRS transformations, meaning all operations with two geometries expect that the geometries have the same CRS.
The Shapely docs don't seem to say much about it though:
ref: https://shapely.readthedocs.io/en/stable/manual.html#coordinate-systems

While it would be technically possible to handle BaseGeometries with different CRSs , it looks like there could be a lot of issues to take into account, and there might not be much demand to provide this capability.
So this limitation is more of a costs vs benefits decision to limit the allowed inputs so we can keep it simple.

bossie (approved):

openeogeotrellis/deploy/batch_job.py
```diff
@@ -120,24 +119,56 @@ def extract_result_metadata(tracer: DryRunDataTracer) -> dict:
...
             geometry = agg_geometry.get_bounding_box_geojson()
             area = agg_geometry.get_bounding_box_area()
+            bbox_crs = agg_geometry.get_crs()
         else:
             logger.warning(f"Result metadata: no bbox/area support for {type(agg_geometry)}")
 
+        # The aggregation geometries return tuples for their bounding box.
+        # Keep the end result consistent and convert it to a list.
+        if isinstance(bbox, tuple):
+            bbox = [*bbox]
```
soxofaan, line 148: I think `bbox = list(bbox)`  is more idiomatic than `[*bbox]`  (I didn't even know this works)

tests/deploy/test_batch_job.py
```diff
@@ -91,6 +93,209 @@ def test_extract_result_metadata_aggregate_spatial_delayed_vector():
...
+    # within the valid bounds of the (projected) CRS of the cube.
+    # That way, we should not get large and/or inaccurate coordinates if any
+    # CRS conversion has happened along the way.
+    #
+    # This smaller rectangle is also the BBox that we expect in the metadata.
+    aggrgeo_west_latlon = 4.2
+    aggrgeo_east_latlon = 4.8
+    aggrgeo_south_latlon = 51.2
+    aggrgeo_north_latlon = 51.8
+    geometries_lat_lon = Polygon(
```
soxofaan, line 266: FYI: shapely has a helper for creating a rectangle like this: `shapely.geometry.box`  https://shapely.readthedocs.io/en/1.8.5/manual.html?highlight=box#shapely.geometry.box

tests/deploy/test_batch_job.py
```diff
@@ -91,6 +93,209 @@ def test_extract_result_metadata_aggregate_spatial_delayed_vector():
...
+            (aggrgeo_east_latlon, aggrgeo_north_latlon),
+            (aggrgeo_east_latlon, aggrgeo_south_latlon),
+            (aggrgeo_west_latlon, aggrgeo_south_latlon),
+        )
+    )
+    # Geojson should always be lat-long, so we don't reproject the the delayed vector file.
+    delayed_vector_file = tmp_path / "delayed_vector_polygon.geojson"
+    geo_json_data = mapping(geometries_lat_lon)
+    with open(delayed_vector_file, "wt") as f:
+        f.writelines(json.dumps(geo_json_data))
```
soxofaan, line 279: FYI: Path objects have write methods to do this more compactly:
```
delayed_vector_file.write_text(json.dumps(geo_json_data))
```

soxofaan (approved):
just some minor notes, nothing blocking

openeogeotrellis/deploy/batch_job.py
JohanKJSchreurs: Done.

tests/deploy/test_batch_job.py
JohanKJSchreurs: Great tip, thanks!
Done.

PR #361, opened 2023-03-06
LINEARRING EMPTY. Add test case with complete graph. Add line to attach java debugger.

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/128


EmileSonneveld:
The added test here is redundant and looks more like an integration test. So no need to merge this.

PR #368, opened 2023-03-10
simplify propagating k8s job name

bossie:
Would this change make sense?

AFAICT a k8s job name can be assembled only once and then persisted in the batch job metadata (`"application_id"`) like in the YARN case, so the propagation is explicit.

At the moment it's being assembled in several places so the propagation is implicit (reconstructed from `user_id` and `job_id`). You could say that it's "by convention" that this works.

This would simplify the code and in particular the job tracker a bit.

Also, semantically speaking I suppose this job name should be unique, like a YARN application ID, no? So in theory we could change it to a UUID and it would still work.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1423,8 +1423,11 @@ class GpsBatchJobs(backend.BatchJobs):
                 jinja_template = Environment(
                     loader=FileSystemLoader(jinja_dir)
                 ).from_string(open(jinja_path).read())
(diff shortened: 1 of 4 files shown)


bossie:
This was merged and seems to work because https://jenkins.vgt.vito.be/job/openEO/job/openeo-deploy-dias/ ran a batch job as part of its pipeline against an https://openeo-dev.creo.vito.be that includes this change.

PR #372, opened 2023-03-23
use polygons to calculate batch job area instead of bbox

JeroenVerstraelen:
https://github.com/Open-EO/openeo-python-driver/issues/144

### DIFF
--- a/tests/test_batch_result.py
+++ b/tests/test_batch_result.py
@@ -1,11 +1,13 @@
 import json
+import pytest
+import xarray
+from openeo.metadata import Band
 from pathlib import Path
 
-import xarray
-import pytest
-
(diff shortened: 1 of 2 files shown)


tests/test_batch_result.py
```diff
@@ -1,11 +1,13 @@
 import json
-from pathlib import Path
-
-import xarray
 import pytest
-
+import xarray
 from openeo.metadata import Band
-from openeogeotrellis.deploy.batch_job import extract_result_metadata,run_job
+from pathlib import Path
```
soxofaan, line 5: I guess this was done automatically with a tool,
 but pathlib is from stdlib, so should not be moved under pytest, xarray, ...

Issue #374, opened 2023-03-24
fix unit test, compare float with pytest.approx


JohanKJSchreurs:
This test fails because there is a very small difference in the comparison of a floating point number:

```
FAILED tests/test_batch_result.py::test_aggregate_spatial_area_result_delayed_vector - assert 187056.0751906881 == 187056.07523286293
```

(This can typically happen on a different machine.)

The easiest is to use [pytest.approx](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx) in the assert.

PR #376, opened 2023-03-24
Issue #374 Add pytest.approx to assert in unit test.

JohanKJSchreurs:
Fixes issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/374

### DIFF
--- a/tests/test_batch_result.py
+++ b/tests/test_batch_result.py
@@ -338,5 +338,5 @@ def test_aggregate_spatial_area_result_delayed_vector(backend_implementation):
     }
     evaluate(pg, env = dry_run_env)
     metadata = extract_result_metadata(dry_run_tracer)
-    assert metadata["area"]["value"] == 187056.07523286293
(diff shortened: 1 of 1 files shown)


JohanKJSchreurs:
Going through the regular flow of creating a GitHub issue + PR even if this is a small fix.

While I could add this into an existing PR soon to be merged, there are already be two different PRs where this go into, and that starts to get a bit messy. This is the clean way to do it.

tests/test_batch_result.py
```diff
@@ -338,5 +338,5 @@ def test_aggregate_spatial_area_result_delayed_vector(backend_implementation):
     }
     evaluate(pg, env = dry_run_env)
     metadata = extract_result_metadata(dry_run_tracer)
-    assert metadata["area"]["value"] == 187056.07523286293
+    assert metadata["area"]["value"] == pytest.approx(187056.07523286293, abs=0.001)
```
soxofaan, line 341: at this scale I would even change the threshold to `abs=0.1` or even coarser. 
Also: no need to keep this number of decimal places, when using `approx` anyway

PR #377, opened 2023-03-27
Dedup error log

EmileSonneveld:
By setting `capture_unhandled_exceptions=False`, but extending try-except in to cover everything


EmileSonneveld:
Best to check "⚙️ > Hide whitespace" when viewing the code

EmileSonneveld:
@bossie when you get back from vacation, do you think it is possible to get rid of `user_facing_logger`? And use the default logging instead? Both end up in Kibana.

bossie:
@EmileSonneveld : I think so; I'll create a ticket with some context.

PR #382, opened 2023-03-30
Add log_level to elasticsearch_log to prevent circuit_breaker_exception

JohanKJSchreurs:
Part of implementation for https://github.com/Open-EO/openeo-python-driver/issues/170

Adding a log level to the ES search will reduce the amount of logs that ES processes so it won't use up so much memory, which is what triggers the circuit breaker.

This fix has to be done across three code bases with their respective PRs

The main PR is: https://github.com/Open-EO/openeo-python-driver/pull/174

The code bases are:

- https://github.com/Open-EO/openeo-geopyspark-driver : Where logs are actually retrieved from Elasticsearch
- https://github.com/Open-EO/openeo-python-driver : needs to link through log_level parameter to backend implementation
- https://github.com/Open-EO/openeo-python-client : needs to be use log_level when retrieving logs via the openeo-python-driver


openeogeotrellis/backend.py
```diff
@@ -2185,13 +2185,19 @@ def get_results_metadata(self, job_id: str, user_id: str) -> dict:
 
         return {}
 
-    def get_log_entries(self, job_id: str, user_id: str, offset: Optional[str] = None) -> Iterable[dict]:
+    def get_log_entries(
+        self,
+        job_id: str,
+        user_id: str,
+        offset: Optional[str] = None,
+        log_level: Optional[str] = None,
```
soxofaan, line 2193: see python driver PR:  will be `level` iso `log_level`

openeogeotrellis/logs.py
```diff
@@ -46,7 +54,7 @@ def elasticsearch_logs(
     """
     try:
         search_after = None if offset in [None, ""] else json.loads(offset)
-        return _elasticsearch_logs(job_id, create_time, search_after)
+        return _elasticsearch_logs(job_id, create_time, search_after, log_level)
```
soxofaan, line 57: I'd start using keyword arguments here

openeogeotrellis/logs.py
```diff
@@ -101,6 +123,11 @@ def _elasticsearch_logs(
                 }
             }
         )
+    if level_filter:
+        query["bool"]["filter"].append(level_filter)
+        print(f"{level_filter=}")
```
soxofaan, line 128: remove the print statements (or replace them with standard `logging.debug` calls)

openeogeotrellis/backend.py
JohanKJSchreurs: Done.

openeogeotrellis/logs.py
JohanKJSchreurs: Done

openeogeotrellis/logs.py
JohanKJSchreurs: Removed them.

PR #394, opened 2023-04-13
Add more logging for STAC projection metatada

JohanKJSchreurs:
Gdalinfo is still receiving a relative path for the asset. Adding more logging to be able to see what is happening.

https://github.com/Open-EO/openeo-geotrellis-extensions/issues/72


JohanKJSchreurs:
It should be fairly safe to merge this once the build on Jenkins passes, so I might just go ahead and merge it.
Nonetheless, I sent some review requests, but it's more to keep people in the know about what is going on.

PR #410, opened 2023-04-27
Auto polarization

EmileSonneveld:
When this is merged, this can be reverted:
https://github.com/Open-EO/openeo-geotrellis-extensions/pull/146


bossie (changes requested):
Added remarks inline.

bossie (approved):
Looks good.

Issue #354, opened 2023-02-28
investigate missing error message for failed card4l sar backscatter job


jdries:
Reported on the forum:
https://discuss.eodc.eu/t/sentinel-1-ard-processing-batch-job-failure-for-long-timeseries/557

This job id:
j-d7c5e0dbd1324640acfdcdb01ee43989

Our logging does not contain a clear failure reason. The last message does indicate that some batch jobs have failed, see below.
It also turned out that the user did not filter on polarization, and this spatiotemporal extent contained GRDH products, which might be problematic.

```

Sentinel
 Hub batch process statuses for batch job 
j-d7c5e0dbd1324640acfdcdb01ee43989: 
{'76f7452d-2e9b-495d-af12-8fe43bdeee25': 'DONE', 
'3fa9718c-e146-4760-8de1-ba9d4c9651fa': 'DONE', 
'4fd28182-101a-4e98-8ca8-3a9435675824': 'DONE', 
'62b9737b-dc6b-4d95-ae6f-58d0b4700c9a': 'DONE', 
'd02d13b3-b93d-4a51-bc9b-6277ea986ed9': 'DONE', 
'0b71c43f-b087-40d6-9524-9e330ff5ab9d': 'DONE', 
'55524e61-e4e1-4584-b5c8-d1575c09d8e1': 'DONE', 
'95314199-f25c-4a10-8b83-98eca9f81bbb': 'PROCESSING', 
'e6b05466-97a1-4c21-a858-dd3ba4f05cd8': 'PROCESSING', 
'3c63609b-27ee-4ca6-95e4-0868731aca01': 'PROCESSING', 
'42b08bd0-117d-4967-a80b-f05e949e6d88': 'PROCESSING', 
'c4a32fe3-57cd-4ca0-82cd-3826fff4650b': 'PROCESSING', 
'27768367-f7e2-4156-a668-571692c2697c': 'PROCESSING', 
'a46c2da1-40cb-427e-93fe-0d5f6e6234fb': 'DONE', 
'25fd791b-5885-4aa3-b058-0774d3b29607': 'PROCESSING', 
'60afe667-3140-4929-8c4d-975789ac5209': 'PROCESSING', 
'e1921187-85ea-4ea0-8231-a765a302101d': 'PROCESSING', 
'38909aaf-bfb1-4a59-b0b3-98f727168b2d': 'PROCESSING', 
[...]

PR #417, opened 2023-05-09
log job failure because of FAILED batch processes + details (if any)

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/354


bossie (commented):
OK.

PR #450, opened 2023-06-14
Issue178 openeo python driver include process graph in result metadata

JohanKJSchreurs:
Implements https://github.com/Open-EO/openeo-python-driver/issues/178

Need an update of both openeoe-python-driver and openeo-geopyspark-driver
See also: https://github.com/Open-EO/openeo-python-driver/pull/202


jdries:
fine for me

soxofaan (approved):

PR #451, opened 2023-06-14
Transform the logging_es_hosts to list to allow definition of multipl…

tcassaert:
…e ES hosts.

### DIFF
--- a/openeogeotrellis/config/config.py
+++ b/openeogeotrellis/config/config.py
@@ -25,7 +25,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
     allow_run_udf_in_driver: bool = False
 
     # TODO: eliminate hardcoded VITO-specific defaults?
-    logging_es_hosts: str = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be")
+    logging_es_hosts: str = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be").split(",")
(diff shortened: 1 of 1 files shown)


openeogeotrellis/config/config.py
```diff
@@ -25,7 +25,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
     allow_run_udf_in_driver: bool = False
 
     # TODO: eliminate hardcoded VITO-specific defaults?
-    logging_es_hosts: str = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be")
+    logging_es_hosts: str = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be").split(",")
```
jdries, line 28: I think the type will have to change from str to list ? Because the split method returns a list of the parts?
May also have impact further on.

openeogeotrellis/config/config.py
```diff
@@ -25,7 +25,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
     allow_run_udf_in_driver: bool = False
 
     # TODO: eliminate hardcoded VITO-specific defaults?
-    logging_es_hosts: str = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be")
+    logging_es_hosts: [str] = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be").split(",")
```
soxofaan:
```suggestion
    logging_es_hosts: List[str] = os.environ.get("LOGGING_ES_HOSTS", "https://es-infra.vgt.vito.be").split(",")
```

Issue #445, opened 2023-06-09
S1 backscatter in 2019: band 20 celltype


jdries:
j-72f8d46bf45f443fb7282bdbb482d60d
Tile id: E514N262 Year: 2019

```
Lost task 0.0 in stage 68.0 (TID 809) (10.42.210.40 executor 5): java.lang.AssertionError: assertion failed: Band 20 cell type does not match, float32 != float64ud0.0
	at scala.Predef$.assert(Predef.scala:223)
	at geotrellis.raster.ArrayMultibandTile.<init>(ArrayMultibandTile.scala:100)
	at geotrellis.raster.ArrayMultibandTile$.apply(ArrayMultibandTile.scala:46)
	at geotrellis.raster.MultibandTile$.apply(MultibandTile.scala:37)
	at org.openeo.geotrellis.OpenEOProcesses$.org$openeo$geotrellis$OpenEOProcesses$$timeseriesForBand(OpenEOProcesses.scala:48)
	at org.openeo.geotrellis.OpenEOProcesses.$anonfun$applyTimeDimension$7(OpenEOProcesses.scala:136)
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158)
	at org.openeo.geotrellis.OpenEOProcesses.$anonfun$applyTimeDimension$3(OpenEOProcesses.scala:134)
	at org.openeo.geotrellis.OpenEOProcesses.$anonfun$applyTimeDimension$14(OpenEOProcesses.scala:153)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:140)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59

```

jdries:
Committed a fix: orfeo was generating float64, while datacube metadata had float32. This is an issue when aggregate_temporal introduces an empty Multiband Tile.

PR #463, opened 2023-06-27
Make delta a bit more lax

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/445


EmileSonneveld:
Fixed here now: https://github.com/Open-EO/openeo-geopyspark-driver/commit/9d700e5ac62c57b5fd6335cc34daf1b8e9a332df

PR #492, opened 2023-08-18
Skipping Terrascope less fast. Use jvm opensearch.

EmileSonneveld:
Sometimes, the code could switch to Creodias, because there where tiles with 95%+ clouds missing. Now, if we don't explicitly ask for them, we do not consider those tiles.

There is now better logging showing what tiles are missing.

Using the JVM opensearch did not give big differences. Only some test mocks needed to be refactored.

### DIFF
--- a/openeogeotrellis/layercatalog.py
+++ b/openeogeotrellis/layercatalog.py
@@ -717,7 +717,7 @@ class GeoPySparkLayerCatalog(CollectionCatalog):
                 )
                 if missing:
                     logger.info(
-                        f"(common_name) {collection_id!r}: skipping {m.provider_backend()!r} because of {len(missing)} missing products."
+                        f"(common_name) {collection_id!r}: skipping {m.provider_backend()!r} because of {len(missing)} missing products: {str_truncate(repr(missing), 1000)}"
                     )
                     continue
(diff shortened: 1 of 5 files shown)


openeogeotrellis/layercatalog.py
```diff
@@ -972,6 +972,36 @@ def _merge_layers_with_common_name(metadata: CatalogDict):
...
+                                          float(_query_kwargs["brx"]), float(_query_kwargs["uly"]))
+    crs = jvm.geotrellis.proj4.CRS.fromEpsgCode(4326)
+    bbox = jvm.geotrellis.vector.ProjectedExtent(extent, crs)
+    attribute_values_dict = {}
+    if "cldPrcnt" in _query_kwargs:
+        attribute_values_dict["eo:cloud_cover"] = _query_kwargs["cldPrcnt"]
+    if "eo:cloud_cover" not in attribute_values_dict:
+        # Terrascope only has <95% clouded tiles, so use as default
+        # So we will not get missing warnings for totally clouded tiles.
+        attribute_values_dict["eo:cloud_cover"] = 95
```
jdries, line 990: actually we cannot do this: if the user does not filter on cloud cover, the default is to include everything, and we should indeed switch to not use Terrascope

EmileSonneveld: removed it

Issue #489, opened 2023-08-10
load_stac: support loading unfinished results


jdries:
Using load_stac, we can normally also load openEO generated results provided by a signed url. 
Using the 'partial' query parameter, it seems possible to get a canonical link to a job that is still running:
https://api.openeo.org/#tag/Data-Processing/operation/list-results
(support for 'partial' needs to be added to our backend)

So if our backend receives a process graph with a load_stac, it should only start the actual processing when all dependencies are finished!
This is an important piece of the puzzle for federated processing, because it allows the aggregator to perform job splitting without keeping track of the various jobs itself: it can just schedule all jobs and forward them to the respective backends. Hence it also no longer needs a long lived token.

bossie:
As discussed: works similarly to SHub batch process polling.

bossie:
"openeo:status" is in the root of the STAC Collection object.

Some considerations/decisions that make sense at this time but might need revisiting:

- since these dependencies are batch jobs and, by definition, can take a bit of time, this flow only makes sense in the context of a batch job (~ SHub batch processes)
- what about a `load_stac(canonical_partial_job_results)` in a `/result` context? Never wait but continue with (partial) results or abort?
- re: the initial check, if "openeo:status" is missing or == "finished", just proceed; if "error" or "canceled" continue with partial results or abort?
- re: the polling check, if "openeo:status" is missing, just proceed; if "error" or "canceled" continue with partial results or abort?

bossie:
Currently implemented as such:

- `load_stac` of partial job results that have already finished with "openeo:status": "error" or "openeo:status": "canceled": just proceed with the batch job; this allows the user to load partial results (their responsibility).
- `load_stac` of unfinished partial job results that eventually fail with "error" or "canceled": fail the the batch job; otherwise the job would proceed with partial results and the user would never know (unless they check the logs)
- `load_stac` of partial job results in `/result`: never wait, proceed with partial results (same reasoning, their responsibility).

bossie:
Did some testing and loading partial job results from OpenEO dev takes a very long time (but not always):

```
[...]

PR #497, opened 2023-08-25
start job: await unfinished job results dependencies #489

bossie:
Implements #489.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1956,6 +1956,10 @@ class GpsBatchJobs(backend.BatchJobs):
             logging_threshold = as_logging_threshold_arg()
 
             def serialize_dependencies() -> str:
+                batch_process_dependencies = [dependency for dependency in
+                                              (dependencies or job_info.get('dependencies') or [])
+                                              if 'collection_id' in dependency]
+
                 def as_arg_element(dependency: dict) -> dict:
                     source_location = (dependency.get('assembled_location')  # cached
                                        or dependency.get('results_location')  # not cached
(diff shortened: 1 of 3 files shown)


openeogeotrellis/backend.py
```diff
@@ -1920,20 +1956,19 @@ def as_logging_threshold_arg() -> str:
             logging_threshold = as_logging_threshold_arg()
 
             def serialize_dependencies() -> str:
-                dependencies = batch_process_dependencies or job_info.get('dependencies') or []
```
bossie, line 1923: Should only serialize SHub batch process dependencies.

We schedule those batch processes ourselves, then pass on their `source_location` (and `card4l` flag) to the batch job that is subsequently started upon their completion ("DONE"). The `load_collection`s that get invoked while evaluating the process graph in the batch job only rely on `source_location` and `card4l`, not on the arguments actually passed to `load_collection`.

In the case of unfinished job results dependencies, however, there's no need to serialize them and pass them on to the subsequent batch job, because `load_stac`s evaluated in that batch job will simply act upon the very same "running" URL that made it a dependency in the first place (but this time it will actually read the job results because they have "finished" in the meanwhile.

openeogeotrellis/backend.py
```diff
@@ -2609,17 +2649,33 @@ def __repr__(self):
...
                         collection_id=collection_id,
                         batch_request_ids=batch_request_ids,  # to poll SHub
                         collecting_folder=collecting_folder,  # temporary cached and new single band tiles, also a flag
                         results_location=f"s3://{bucket_name}/{subfolder}",  # new multiband tiles
                         card4l=card4l  # should the batch job expect CARD4L metadata?
                     ))
+            elif process == 'load_stac':
+                url, _ = arguments  # properties will be taken care of @ process graph evaluation time
+
+                with requests_with_retry().get(url) as stac_resp:
```
bossie, line 2662: Share an existing Session instead of creating a new one.

Issue #467, opened 2023-06-30
support & example of running backend as local docker container


jdries:
For EOEPCA docker & HPC runners:

Two options:
* start webapp locally
* start job based on process graph json, so command line tool, no webapp

Use load_stac to load data

Startpoint:
https://github.com/Open-EO/openeo-geotrellis-kubernetes/blob/master/examples/run_local.md

Document everything + simplify where possible.

Epic: EOEPCA/roadmap#10

soxofaan:
FYI: long time ago I already made a command line tool to execute a process graph with geopyspark driver machinery:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/19740a43e75eb49c7461f33bcf7b35f7efdc69f1/openeogeotrellis/cli.py#L1-L17

It's pretty old, so I'm not sure if it still works

jdries:
This one needs to see how we enable 'in-memory' job registry.

JohanKJSchreurs:
Brief summary from PR https://github.com/Open-EO/openeo-geopyspark-driver/pull/533
- [x]  Be able to run test suite (pytest) in Docker.
- [x]  Running backend / web app as local docker container, namely be able to use the REST API.
- [ ]  Be able to run a batch job for testing.   --> This needs the in-memory job register to be able to track the batch jobs

JeroenVerstraelen:
This issue continues in the following EPIC:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/806

PR #533, opened 2023-10-09
Issue467 part2 running backend as local docker container

JohanKJSchreurs:
Part 2 of issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/467 Add initial support for developing with local docker container

Making it possible to run the backend web app / rest API
Should also be possible to run batch jobs for testing, for example via `openeogeotrellis/cli.py`


JohanKJSchreurs:
Going to merge this PR to make the debugging support in Docker available on other branches.

This should be safe because:
-  It doesn't change the code itself, only a separate Docker setup for development.
-  You have to enable this setup explicitly to us it, i.e. build the Docker image an run it.

PR #584, opened 2023-11-13
Creo missing products

EmileSonneveld:


tests/test_api_result.py
```diff
@@ -75,6 +75,13 @@ def set_all(properties: dict):
         set_all(orig_properties)
 
 
+@pytest.fixture
+def catalog(vault):
```
soxofaan, line 79: why is this fixture added? It doesn't seem to be used

tests/test_api_result.py
```diff
@@ -42,15 +42,15 @@
...
 from openeogeotrellis.testing import random_name, KazooClientMock, config_overrides
 from openeogeotrellis.utils import (
     UtcNowClock,
     drop_empty_from_aggregate_polygon_result,
     get_jvm,
     is_package_available,
 )
-from openeogeotrellis.vault import OAuthCredentials
-from .data import get_test_data_file
+from tests.data import get_test_data_file
```
soxofaan, line 53: I'm not sure this is an improvement over the original `from .data import get_test_data_file` (which is also not ideal by the way). This import makes it look like there is some top-level `tests` package in the venv while the relative import makes it more clear it is just from a sibling file.

Is there a particular reason you changed this?

soxofaan (commented):
some minor notes

PR #585, opened 2023-11-14
Remove broken missing products check to avoid accidentally using it. 

EmileSonneveld:
Replaces this PR: https://github.com/Open-EO/openeo-geopyspark-driver/pull/584

### DIFF
--- a/openeogeotrellis/layercatalog.py
+++ b/openeogeotrellis/layercatalog.py
@@ -1045,8 +1045,9 @@ def check_missing_products(
 
         method = check_data.get("method")
         if method == "creo":
-            creo_catalog = CreoCatalogClient(**check_data["creo_catalog"])
-            missing = [p.getProductId() for p in creo_catalog.query_offline(**query_kwargs)]
+            logger.warning("At the moment it is not supported to check missing products on creo.")
+            # We could query with status="all" and see if some offline products survive the deduplication and log those.
+            missing = []
         elif method == "terrascope":
             jvm = get_jvm()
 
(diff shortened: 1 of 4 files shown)


openeogeotrellis/layercatalog.py
```diff
@@ -1045,8 +1045,9 @@ def check_missing_products(
 
         method = check_data.get("method")
         if method == "creo":
-            creo_catalog = CreoCatalogClient(**check_data["creo_catalog"])
-            missing = [p.getProductId() for p in creo_catalog.query_offline(**query_kwargs)]
```
soxofaan, line 1049: maybe also put a TODO comment in the `query_offline` explaining why the implementation needs work

EmileSonneveld:
@soxofaan Can this check be removed then? https://github.com/Open-EO/openeo-geopyspark-driver/commit/c63728c3cad9fb71a63bb031d6561ad71880e3c1#diff-e48ae7baaf533de3301112527309f195fef859e933e1e17a1955b3bc833af33aR1513

openeogeotrellis/layercatalog.py
EmileSonneveld: Added a comment

soxofaan:
indeed, that if from c63728c3cad9fb71a6  can also be removed with this PR

EmileSonneveld:
So, the broken missing products check is removed, the test for that code too, and `extra_validation` could be re-enabled. @soxofaan , this is good to merge, right?

Issue #531, opened 2023-10-06
Support multiple accounting backends in aggregator scenario


JanssenBrm:
With the deployment of the aggregator, processing can be divided on multiple back-ends. For example, when deploying the aggregator in CDSE, some of the processing could be done on the TerraScope environment. 

In order to fully support this, the final processing environment (in the example above, TerraScope) should interact with the proper accounting backend (in the example, the CDSE marketplace-cost-api). Specifically, whenever TerraScope gets a request to process a job from CDSE it should:

- Execute the credit check against the CDSE backend
- Report the resources used to the CDSE backend

Some possible approaches for openEO to determine the correct accounting backend have already been discussed with @soxofaan :
- A user agent could identify from which environment the request is coming.
- Inspection of the token could achieve the same thing.

soxofaan:
related: https://github.com/eu-cdse/openeo-cdse-infra/issues/27

soxofaan:
did first successful experiment today:
created job (j-240208eceedf489dbe7ad113d25e0bf0) on "openeo-dev.vito.be" backend with `job_options={"etl_api_id": "CDSE-staging"}` and the job tracker (for dev) sent the resource usage data to the CDSE staging ETL API

soxofaan:
setup was done through `openeogeotrellis.integrations.etl_api.MultiEtlApiConfig` usage in `GpsBackendConfig` in openeo-deploy project

soxofaan:
I'm going to close this ticket for now. 
There is still other work to be done (e.g. at level of aggregator: https://github.com/Open-EO/openeo-aggregator/issues/135, and aligning OIDC providers in multiple deployments), but that is food for other tickets

PR #589, opened 2023-11-20
Simplify test job tracker v2.py

bossie:
Adds test for https://github.com/eu-cdse/openeo-cdse-infra/issues/35.


tests/test_job_tracker_v2.py
```diff
@@ -356,6 +316,20 @@ def submit(
         return app
 
 
+class PrometheusMock:
```
bossie, line 319: Replace with `mock.Mock(Prometheus)`.

See `tests.test_job_tracker_v2.TestK8sStatusGetter.test_cpu_and_memory_usage_not_in_prometheus`.

bossie:
A bit of context: adding a test for the `K8sStatusGetter`, where the Prometheus API has no records for the CPU/memory usage of a batch job, in which case:

- a warning is issued so we are aware of this situation,
- default values for CPU/memory are reported instead.

To simulate Prometheus returning no values for CPU/memory usage, I found adapting `KubernetesMock` complicated so I simplified things a bit. It essentially comes down to avoid patching module level functions like `openeogeotrellis.job_tracker_v2.kube_client` and `requests.get` and providing mock objects instead; it's still mocking but on a higher level.

Passing `kube_client` and `Prometheus` dependencies explicitly to `K8sStatusGetter` instead of having them buried deeply in the implementation is also a plus:

https://github.com/Open-EO/openeo-geopyspark-driver/blob/f4887244ab78da0b7b16b3bc45ffbb1512681969/openeogeotrellis/job_tracker_v2.py#L594-L595

bossie:
Mostly test code, easy to revert so merged.

soxofaan:
yeah sorry, I had this on my todo list for a long time but never got to it  :smirk:

I also did some tweaking to the job tracker v2 utilities in the meantime to support #531  FYI

Issue #203, opened 2022-08-18
Better error for user when requesting too much batch job memory or cpu


soxofaan:
from https://discuss.eodc.eu/t/memory-overhead-problem/424/9

user tried submitting job with a lot of requested memory:
```
job_options = {
        "executor-memory": "100G",
        "executor-memoryOverhead": "100G",
        "executor-cores": "4"}
```

but that failed. Error message/log for user were just a cryptic 
```
No such file or directory: '/data/projects/OpenEO/j-3c44cd000ba648a8b6350a0765d8e48c/log
```
(with a lot of cruft around it).

I had to look at the logs emitted at https://github.com/Open-EO/openeo-geopyspark-driver/blob/60f9abcc7e187b4d731b82e2014a875c83a9aaf0/openeogeotrellis/backend.py#L1268

to find well hidden

> Exception in thread "main" java.lang.IllegalArgumentException: Required executor memory (102400 MB), offHeap memory (0) MB, overhead (102400 MB), and PySpark memory (0 MB) is above the max threshold (52224 MB) of this cluster! Please check the values of 'yarn.scheduler.maximum-allocation-mb' and/or 'yarn.nodemanager.resource.memory-mb'.

We should improve the error message and logging for the end user here

jdries:
situation on CDSE: max cpu is 4, if user requests more, the job hangs

tcassaert:
I think even requesting 4 won't work, as there's never the full 4 cpus available.

EmileSonneveld:
Setting `'driver-cores': '5'` can also make the job hang. A high `executor-cores` shows a user friendly error.

EmileSonneveld:
This is deployed on https://openeo-staging.creo.vito.be:
Example message: `openeo.rest.OpenEoApiError: [400] Internal: Requested too many driver cores: 19 , the max for this instance is: 4 (ref: r-231204b8b72b4fae92a153858dfa08cf)`

EmileSonneveld:
@tcassaert `'executor-cores': '4'` works 4, but `'driver-cores': '4'` indeed not. What is the reason? Is the max for a driver `maxCores - 1`?

EmileSonneveld:
As discussed with @tcassaert, it is unclear why `'executor-cores': '4'` works. (example: `j-231204a7816f41518f5aa2336f364719`) So we can keep the limit on 4 for the moment.

PR #598, opened 2023-12-04
Also error for too much driver memory/cores. 

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/203

### DIFF
--- a/openeogeotrellis/config/config.py
+++ b/openeogeotrellis/config/config.py
@@ -80,7 +80,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
 
     prometheus_api: Optional[str] = os.environ.get("OPENEO_PROMETHEUS_API")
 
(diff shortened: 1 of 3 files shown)


openeogeotrellis/config/config.py
```diff
@@ -80,7 +80,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
 
     prometheus_api: Optional[str] = os.environ.get("OPENEO_PROMETHEUS_API")
 
-    max_executor_memory: str = "64G"
+    max_executor_or_driver_memory: str = "64G"  # Executors and drivers have the same amount of memory
```
EmileSonneveld, line 83: @jdries Is this rename ok?
As I understood, driver and executor often have the same specs

jdries: sure!

Issue #392, opened 2023-04-11
`load_results`/`load_stac`: support STAC items


soxofaan:
Current `load_results` implementation requires that the provided result URL (metadata):
-  is a valid STAC Collection 
- uses signed URLs when linking to assets (URLs that load without necessary auth headers).

The openEO API spec however also allows batch job results to be STAC items (type "Feature" iso "Collection).

Also, it should be possible to use non-signed batch job result URLs of the same user (e.g. a user with batch job `j-123` should be able to use `load_result("https://openeo../jobs/j-123/results")` directly, without need for canonical/signed URLs)

soxofaan:
related use case: https://github.com/openEOPlatform/architecture-docs/issues/316

soxofaan:
STAC item support has been added to load_stac by now

not sure what current state is concerting signed or non-signed batch job result URLs

bossie:
If the batch job results document is a STAC Collection, like it is in the VITO implementation, an unsigned URL is supported; if the results document is a STAC Item, it might not be (untested).

I did however recently observe that pointing `load_stac` to a signed child Item of a job results Collection fails: https://github.com/Open-EO/openeo-geopyspark-driver/issues/614

PR #613, opened 2023-12-14
Support load_stac on item without "eo:bands" under properties

soxofaan:
get band names directly from STAC Item assets, instead of Item's "eo:bands" property (which might not exist) or parent collection metadata (which might not be present)

(related to #392)


bossie:
The lack of Item `eo:bands` is not the problem; the problem is rather that if there is no Item `eo:bands`, it tries to fall back to Collection `eo:bands` but this Item is not linked to a Collection so it fails.

The band _names_ absolutely come from the assets themselves; it's the band _order_ that it will try to derive from Item/Collection `eo:bands`. If there are no such `eo:bands`, the order will be unpredictable (but nevertheless the same for all Items considered).

I think that, with your implementation, an Item that _doesn't_ have an `eo:bands` but _is_ linked to a Collection with an `eo:bands`, will not conform to the order of its Collection `eo:bands`.

I also think that just replacing it with:

```python
if "eo:bands" in item.properties:
    eo_bands_location = item.properties
elif item.get_collection() is not None:
    eo_bands_location = item.get_collection().summaries.lists
else:
    eo_bands_location = {}
```

will fix the original problem and apply the proper order as well.

bossie:
I'll add a test case.

Issue #568, opened 2023-11-06
load_stac raises ValueError when STAC file uses "Z" as TZ for UTC timestamps


JohanKJSchreurs:
## load_stac raises ValueError when STAC file uses "Z" as TZ for UTC timestamps

`connection.load_stac(...)` raises a ValueError when the UTC timestamps in the STAC file use the TZ format with "Z", for example: '2020-01-01T00:00:00Z'
This format is commonly used, so it is best that we support it too.

This happens with Python 3.8 because `datetime.datetime.fromisoformat` did not yet support this format yet in Python 3.8.
Though, it is supported starting from Python 3.11.
See also: https://docs.python.org/3/library/datetime.html?highlight=datetime%20fromisoformat#datetime.datetime.fromisoformat

However, we can catch this inside the load_stac implementation and handle it there. 

### Exception info

Error on the client side:
openeo.rest.OpenEoApiError: [500] Internal: Server error: ValueError("Invalid isoformat string: '2020-01-01T00:00:00Z'") 

Error traceback from server logs:
```
"Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.8/site-packages/openeo_driver/users/auth.py", line 88, in decorated
    return f(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/openeo_driver/views.py", line 631, in result
    result = backend_implementation.processing.evaluate(process_graph=process_graph, env=env)
  File "/usr/local/lib/python3.8/site-packages/openeo_driver/ProcessGraphDeserializer.py", line 300, in evaluate
    return evaluate(process_graph=process_graph, env=env)
  File "/usr/local/lib/python3.8/site-packages/openeo_driver/ProcessGraphDeserializer.py", line 370, in evaluate
    result = convert_node(result_node, env=env)
  File "/usr/local/lib/python3.8/site-packages/openeo_driver/ProcessGraphDeserializer.py", line 390, in convert_node
    process_result = apply_process(process_id=process_id, args=processGraph.get('arguments', {}),
[...]

PR #615, opened 2023-12-14
Issue568 fix loadstac raises valueerror when utc timestamps have z as tz

JohanKJSchreurs:
## Issue568 fix loadstac raises valueerror when utc timestamps have z as tz 'Z': 2020-01-01T00:00:00Z

Issue: [#568](https://github.com/Open-EO/openeo-geopyspark-driver/issues/568)

PR currently in draft:
- Found a fix but haven't implemented proper test coverage yet.
- Started on the test but it is not working yet. openeo can't find the STAC file it seems. Therefore: `pytest.mark.skip`
- Still want to share it, and possible get some feedback already.

Looks like this test does need to read the STAC collection from a file. This regression test seems harder to do with mocks, because mocks typically contain the "parsed" pystac objects, not the raw data from the file. And this bug is really about different timezone formats that are allowed but that hit an edge case: Python 3.8 did not fully support the correct ISO formats yet.

An alternative solution could be to upgrade to Python 3.11 but we have to be careful when we increase the required Python version.
If we would do that, every other implementation of the openeo-python-driver may have to support 3.11. Not sure if that is a good idea.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -785,10 +785,28 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
         temporal_extent = load_params.temporal_extent
         from_date, to_date = normalize_temporal_extent(temporal_extent)
 
+        def from_isoformat(datetime_str: str) -> dt.datetime:
+            """Fix for Python <3.11:
+            dt.datetime.fromisoformat does not support the format with "Z" at
+            the end to denote the time zone is UTC (Zulu time).
+            e.g.: "2020-03-17T12:34:56Z"
+            This helper function can be removed when we upgrade to python 3.11,
+            but in 3.8 we need to handle this format ourselves.
+            """
+            try:
+                return dt.datetime.fromisoformat(datetime_str)
+            except:
+                # Fall back to our own datetime format handler.
+                # However, this one does not like the other common format: "+00:00"
+                # otherwise we could just use rfc3339.
+                # TODO: [openeo-python-client] support "+00:00" in Rfc3339.
+                #   In fact, there is already a TODO for supporting other time zones.
+                return rfc3339.parse_datetime(datetime_str, with_timezone=True)
+
         def intersects_spatiotemporally(itm: pystac.Item) -> bool:
             def intersects_temporally() -> bool:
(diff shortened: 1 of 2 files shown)


JohanKJSchreurs:
Though the PR is still in draft because this is work in progress, I requested reviewers who are working on related issues, to keep them in the know.

Nonetheless, @soxofaan and @bossie if you happen to have feedback don't hesitate to let me know.

bossie (changes requested):
With a small change I managed to get the test running: https://github.com/Open-EO/openeo-geopyspark-driver/commit/bcfcec9466872daa2a327c1ddb7acf9a7f743e75

The problem is that the test runs successfully even without the fix; this is because the issue lies in the parsing of `start_datetime` and this STAC Item doesn't have one (it has a regular `datetime`, which pystac manages to parse successfully).

Instead of creating the STAC Collection in code and saving it to disk, I would just do like the other tests and add regular JSON files to the repo. That way:
- it makes it immediately clear what your STAC looks like;
- you don't depend on the way pystac renders your UTC timestamps in the first place (does it actually render the time zone as 'Z' or e.g. '+00:00'?).

For an example STAC Item with `start_datetime`/`end_datetime` you might want to copy an existing one and adapt e.g. from https://vlcc.geoville.com/resto/collections/BVL_v1/items; this particular collection is conveniently one that we actually want to support.

As for the fix itself: instead of first trying with `datetime.fromisoformat`, then falling back to `rfc3339.parse`, maybe try a more lenient parser that supports both 'Z' and timezone offsets like `dateutil.parser.parse` (python-dateutil is already a dependency).

soxofaan:
> As for the fix itself: instead of first trying with datetime.fromisoformat, then falling back to rfc3339.parse, maybe try a more lenient parser that supports both 'Z' and timezone offsets like dateutil.parser.parse (python-dateutil is already a dependency).

Indeed, I would also prefer a generic solution like that over adhoc and version dependent try-except constructs.
Also note that rfc3339 from the python client is primarily intended as a best effort date (re)formatter, while super robust date parsing is not its main focus.

JohanKJSchreurs:
Closing this PR: it is no longer needed because @bossie was able to fix this already in https://github.com/Open-EO/openeo-geopyspark-driver/pull/673.

Issue #632, opened 2024-01-04
phase out ZkJobRegistry


bossie:
Context: ZK on CDSE crashes due to OOM.

soxofaan:
FYI: related to (and maybe even duplicate of) #498

bossie:
`async_task` is only aware of `ZkJobRegistry`.

In the case of SHub batch processes that's just fine because those weren't/aren't available on CDSE SHub endpoint https://sh.dataspace.copernicus.eu and won't be considered:

https://github.com/Open-EO/openeo-geopyspark-driver/blob/4478f96a3053eda10851cf4afd5cc2dd657e1704/openeogeotrellis/backend.py#L2577-L2578

`load_stac` of partial job results only works on Terrascope anyway so this changes nothing.

soxofaan:
FYI: I've also been tagging some implementation parts with todo notes referencing `#498`
```
$ grep '#498' openeogeotrellis/ -r -n
openeogeotrellis/job_registry.py:797:        # TODO #236/#498 For now: compare job metadata between Zk and EJR
openeogeotrellis/job_registry.py:910:        # TODO #236/#498: error if both sources failed?
openeogeotrellis/job_registry.py:923:        # TODO #236/#498 Need to have EJR implementation for this? This is only necessary for ZK cleaner script anyway.
openeogeotrellis/job_tracker_v2.py:395:            # TODO: #236/#498 also/instead get jobs_to_track from EJR?
openeogeotrellis/backend.py:334:            # TODO #236/#498 avoid this fallback and just make sure it is always set when necessary
openeogeotrellis/backend.py:1640:            zk_job_registry_factory=ZkJobRegistry,  # TODO #236/#498 allow to disable this with config?
openeogeotrellis/backend.py:1984:            # TODO #498 eliminate ZK code path, or at least encapsulate this logic better
```

bossie:
Heh. Persisting usage in ES (#488) doesn't work. From the YARN job_tracker logs:

> In context "job_tracker job_metadata.status='finished' from YarnStatusGetter": caught EjrHttpError('EJR API error: 400 \'Bad Request\' on `PATCH \'https://jobregistry.openeo.vito.be/jobs/j-240105907a9640a79375dbcdabe66505\'`: {"statusCode":400,"message":["property costs should not exist"],"error":"Bad Request"}')

```
Traceback (most recent call last):
[...]

PR #635, opened 2024-01-09
632 phase out zkjobregistry

bossie:
Implements #632.

PR #639, opened 2024-01-15
quickfix: avoid hanging ZK get_job calls #632

bossie:
Context: calls to get a batch job from ZK seem to hang and occupy Gunicorn threads, blocking other requests.

The quickfix is to favor EJR and thus relieve ZK in this case.


jdries (approved):
LGTM!

bossie:
This was ultimately rolled back because it messed up the batch job result metadata (in particular, its spatial and temporal extent). #635 should fix this in a more general way.

PR #641, opened 2024-01-16
Block too large sync job

EmileSonneveld:
Implemented in is_layer_too_large. Also do this pixel volume check on all sync jobs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     tests_require=tests_require,
     install_requires=[
         "openeo>=0.15.0",
-        "openeo_driver==0.80.0a1_block_large_requests",
+        "openeo_driver>=0.78.0.dev",
         'pyspark==3.3.1; python_version>="3.8"',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@
     tests_require=tests_require,
     install_requires=[
         "openeo>=0.15.0",
-        "openeo_driver>=0.78.0.dev",
+        "openeo_driver==0.80.0a1_block_large_requests",
```
soxofaan, line 56: I don't think this is something we want to merge in master

soxofaan (changes requested):

setup.py
EmileSonneveld: This will be removed during merging. It was an attempt to not interfere with the main branch.
Actually I can remove it now, as it did not work...

jdries (approved):

setup.py
EmileSonneveld: FFR: https://peps.python.org/pep-0440/#pre-release-separators

PR #660, opened 2024-01-29
632 revert to zk primary job registry

bossie:
OpenEO and Job Registry Elastic API are in better shape but issues beyond our control i.e. dropped connections have too much of an impact. This PR makes `ZkJobRegistry` the primary job data store again.


openeogeotrellis/job_registry.py
```diff
@@ -946,13 +969,20 @@ def get_all_jobs_before(
...
+        ejr_jobs = None
+
         if self.zk_job_registry:
-            yield from self.zk_job_registry.get_running_jobs(parse_specification=True)
-        elif self.elastic_job_registry:
-            yield from self.elastic_job_registry.list_trackable_jobs(fields=[
-                "job_id", "user_id", "application_id", "status", "created", "title", "job_options", "dependencies",
-                "dependency_usage",
-            ])
+            zk_jobs = [j for j in self.zk_job_registry.get_running_jobs(parse_specification=True)]
```
bossie: Can be replaced with `list()`.

PR #661, opened 2024-01-29
Add skipped test_layer_catalog_step_resolution. 

EmileSonneveld:
Better temporal and spatial resolution parsing


tests/test_layercatalog.py
```diff
@@ -231,6 +231,59 @@ def test_get_layer_catalog_opensearch_enrich_creodias(requests_mock, vault):
...
+        native_resolution = {
+            "cell_width": cell_width,
+            "cell_height": cell_height,
+            "crs": crs,
+            # "crs": "EPSG:4326",
+        }
+        reprojected = reproject_cellsize(spatial_extent, native_resolution, "Auto42001")
+        print(f"Resolution in meters: {reprojected}")
+        if (not 0.01 < reprojected[0] < 100000) or (not 0.01 < reprojected[1] < 100000):
+            print("WARNING: reprojected cellsize in meters is not in expected range: " + str(reprojected[0]) + "m")
```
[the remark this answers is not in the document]
EmileSonneveld: I added all warnings to a string and asserted it is empty at the end. Good fix?

EmileSonneveld:
I used the snippet to find bad "step" metadata and fix it here: https://git.vito.be/projects/BIGGEO/repos/openeo-deploy/pull-requests/14/overview

tests/test_layercatalog.py
```diff
@@ -231,6 +231,59 @@ def test_get_layer_catalog_opensearch_enrich_creodias(requests_mock, vault):
     ]
 
 
+@pytest.skip("Run manually when changing layercatalog.json files")
+def test_layer_catalog_step_resolution(vault):
+    with mock.patch("openeogeotrellis.layercatalog.ConfigParams") as ConfigParams:
+        ConfigParams.return_value.layer_catalog_metadata_files = [
+            "/home/emile/openeo/openeo-deploy/mep/layercatalog.json",
```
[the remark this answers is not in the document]
soxofaan: I think it makes sense to move the logic to a utility function that takes a catalog file path as argument.
We can then call that function from integration tests in the appropriate repo's (e.g. openeo-deploy, etc).

openeogeotrellis/geopysparkcubemetadata.py
```diff
@@ -0,0 +1,153 @@
...
+from typing import List, Union
+
+from openeo.metadata import CollectionMetadata, Dimension
+from openeogeotrellis.utils import reproject_cellsize
+
+_log = logging.getLogger(__name__)
+
+
+def clean_2D_tuple(tuple_to_clean):
+    if not tuple_to_clean:
```
soxofaan, line 11: what is 2D about the returned value?

openeogeotrellis/geopysparkcubemetadata.py
```diff
@@ -0,0 +1,153 @@
...
+        return None
+    if isinstance(tuple_to_clean, float) or isinstance(tuple_to_clean, int):
+        return tuple_to_clean, tuple_to_clean
+    if (tuple_to_clean[0] is None) or (tuple_to_clean[1] is None):
+        return None
+    if isinstance(tuple_to_clean, tuple):
```
soxofaan, line 17: why no length check here?

tests/test_layercatalog.py
```diff
@@ -231,6 +234,158 @@ def test_get_layer_catalog_opensearch_enrich_creodias(requests_mock, vault):
...
+            print(f"{temporal_step_str=}")
+
+        if metadata.get("id") == 'WATER_BODIES':
+            print("ha")
+        gsd_in_meter = metadata.get_GSD_in_meters()
+        print(f"{gsd_in_meter=}")
+        datasource_type = metadata.get("_vito", "data_source", "type")
+        warn_str = layer["id"] + " datasource_type:" + str(datasource_type) + "\n"
+
+        def clean_2D_tuple(tuple_to_clean):
```
soxofaan, line 269: is it intentional to duplicate this clean_2D_tuple from geopysparkcubemetadata.py?

openeogeotrellis/utils.py
```diff
@@ -572,3 +574,85 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     return any(m.name == name for m in pkgutil.iter_modules())
+
+
+def reproject_cellsize(
+        spatial_extent: dict,
+        native_resolution: dict,  # cell_width, cell_height, crs
+        to_crs: str,
+) -> Tuple[float, float]:
+    """
+    :param spatial_extent: Cellsize conversion could be different when done at the poles or the equator
```
soxofaan, line 585: this param description is a bit confusing

openeogeotrellis/utils.py
```diff
@@ -572,3 +574,85 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 @functools.lru_cache
 def is_package_available(name: str) -> bool:
     return any(m.name == name for m in pkgutil.iter_modules())
+
+
+def reproject_cellsize(
```
soxofaan, line 579: this function deserves a couple of unit tests I think

soxofaan (commented):
some quick notes

tests/test_layercatalog.py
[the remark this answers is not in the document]
EmileSonneveld: I changed the path to a layercatalog in the same project that has many of the same layers

openeogeotrellis/geopysparkcubemetadata.py
EmileSonneveld: Comment added: "Convert input an (x,y) tuple if possible."

openeogeotrellis/geopysparkcubemetadata.py
EmileSonneveld: true, fixed

tests/test_layercatalog.py
EmileSonneveld: no, deduped. Thanks!

openeogeotrellis/utils.py
EmileSonneveld: true, added under `test_reproject_cellsize(...)`

openeogeotrellis/utils.py
EmileSonneveld: Is this ok?
 ```
The spatial extent is needed, because conversion is often
     different when done at the poles compared to the equator.
     eg: When converting 1meter to degrees (in LatLon) at the North Pole, it can be 14% degrees more in LatLon
     compared to the same conversion at the equator.
```

File: openeogeotrellis/geopysparkcubemetadata.py (hunk 3 above)
Comment (soxofaan) [reply openeogeotrellis/geopysparkcubemetadata.py:14]: no I mean, (x,y) is not a 2D tuple, it is a 1D tuple with 2 items

File: tests/test_utils.py [hunk 8]
```diff
@@ -267,3 +267,81 @@ def test_to_s3_url(file_or_folder_path, bucket_name, expected_url, monkeypatch):
...
+                {'cell_height': 10, 'cell_width': 10, 'crs': 'EPSG:32632'},
+                'EPSG:4326',
+                (0.0000887560370977725, 0.00008935420776900408)
+        ),
+    ],
+)
+def test_reproject_cellsize(spatial_extent, native_resolution, to_crs, expected):
+    projected_resolution = reproject_cellsize(spatial_extent, native_resolution, to_crs)
+    print(projected_resolution)
+    assert abs(projected_resolution[0] - expected[0]) < 1e-7
```
soxofaan, line 321: FYI  pytest has a feature to avoid this boilerplate with `print` and improve the error message when it fails:

```python
assert projected_resolution[0] == pytest.approx(expected[0], abs=1e-7)
```

Or even better, check pair as a whole for even more insights when things go wrong

```python
assert projected_resolution == tuple(pytest.approx(x, abs=1e-7) for x in expected)
```

openeogeotrellis/utils.py
```diff
@@ -572,3 +574,89 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        spatial_extent: dict,
+        native_resolution: dict,  # cell_width, cell_height, crs
```
soxofaan, line 581: Without more context, it is not clear what "native" means here. This function is just about converting cell size, but there are no assumptions about what is native. I would just call this  `cell_size`.

also note that the current API is a bit asymmetric: the input cell size should be a dictionary, but the output cell size is a tuple

openeogeotrellis/utils.py
soxofaan: I think it can even be more than 14%:  the longitude degrees per meter ratio goes to infinity at the poles

openeogeotrellis/geopysparkcubemetadata.py
soxofaan: yeah "tuple2" or "pair" make more sense I think

openeogeotrellis/utils.py
EmileSonneveld: I split the `native_resolution` into `input_resolution` and `input_crs`

openeogeotrellis/geopysparkcubemetadata.py
EmileSonneveld: clean_number_pair

openeogeotrellis/utils.py
EmileSonneveld: I made it more general. When I did the calculation, I checked how much the Y size changed. The X-size seems to change 9x when going at the north side. However, the UTM zone I used to check only went till 85deg to the north pole

tests/test_utils.py
EmileSonneveld: nice. Used

soxofaan (approved):
I gave it a quick last review and looks fine to me

PR #673, opened 2024-02-05
load_stac from netCDF

### LINKED ISSUE #646: Batch job results don't contain any items when saved as netCDF
When saving results of a batch job as netCDF, the resulting STAC collection doesn't contain any items and therefore cannot be loaded using `load_stac`.

For example the following batch job:
[j-2401190a5a2144868480ccba676ee9db.json](https://github.com/Open-EO/openeo-geopyspark-driver/files/14010267/j-2401190a5a2144868480ccba676ee9db.json)

will result in the following STAC metadata:
[job-results.json](https://github.com/Open-EO/openeo-geopyspark-driver/files/14010271/job-results.json)

Using `load_stac` on these results will result in a NoDataAvailable Exception.

bossie:
Fixes #646.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -21,6 +21,7 @@ from subprocess import CalledProcessError
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
+import dateutil
 import flask
 import geopyspark as gps
 import kazoo.exceptions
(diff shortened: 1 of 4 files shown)


bossie:
Also fixes #568.

bossie:
Running the whole chain locally became exceedingly difficult so merged this in early. It persists the netCDF assets' extents and geometries as they will later become those of the corresponding the STAC Item.

PR #679, opened 2024-02-06
Oscars pagination

EmileSonneveld:
Bigger page size, check multiple pages, dedup for when running in auto tests. Only check next page when needed

### DIFF
--- a/openeogeotrellis/opensearch.py
+++ b/openeogeotrellis/opensearch.py
@@ -25,7 +25,7 @@ class OpenSearchOscars(OpenSearch):
             cache = {}
             start_index = 1
             while True:
-                url = f"{self.endpoint}/collections?startIndex={start_index}&count=200&sortKeys=start"
+                url = f"{self.endpoint}/collections?startIndex={start_index}&count=200"
                 logger.info(f"Getting collection metadata from {url}")
                 resp = requests.get(url=url)
                 resp.raise_for_status()
@@ -41,7 +41,7 @@ class OpenSearchOscars(OpenSearch):
                     # Or could be because we reached the end of the list.
                     break
                 if len(cache) != cache_length_before + len(collection):
-                    raise Exception(
+                    logger.warning(
                         f"Expected {len(collection)} features to be added for this page, but got {len(cache) - cache_length_before}")
                 # We can't break the loop early using 'totalResults' as has shown to be unreliable.
                 start_index += len(collection)
--- a/openeogeotrellis/opensearch.py
+++ b/openeogeotrellis/opensearch.py
@@ -32,21 +32,21 @@ class OpenSearchOscars(OpenSearch):
                 resp = requests.get(url=url)
                 resp.raise_for_status()
                 json = resp.json()
-                collection = json["features"]
+                collections = json["features"]
                 cache_length_before = len(cache)
-                for f in collection:
-                    cache[f["id"]] = f
+                for collection in collections:
+                    cache[collection["id"]] = collection
 
                 # If nothing got added, we're done.
                 if len(cache) == cache_length_before:
                     # could be because in auto-tests each page contains the same features.
                     # Or could be because we reached the end of the list.
                     break
-                if len(cache) != cache_length_before + len(collection):
+                if len(cache) != cache_length_before + len(collections):
                     logger.warning(
-                        f"Expected {len(collection)} features to be added for this page, but got {len(cache) - cache_length_before}")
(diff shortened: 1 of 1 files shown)


openeogeotrellis/opensearch.py
```diff
@@ -22,11 +22,29 @@ def __init__(self, endpoint: str):
...
-            self._cache = {f["id"]: f for f in resp.json()["features"]}
+            cache = {}
+            start_index = 1
+            while True:
+                url = f"{self.endpoint}/collections?startIndex={start_index}&count=200"
+                logger.info(f"Getting collection metadata from {url}")
+                resp = requests.get(url=url)
+                resp.raise_for_status()
+                json = resp.json()
+                features = json["features"]
```
bossie, line 33: Rename variable to `collections`.

openeogeotrellis/opensearch.py
```diff
@@ -22,11 +22,29 @@ def __init__(self, endpoint: str):
...
+                cache_length_before = len(cache)
+                for f in features:
```
bossie, line 35: Rename variable to `collection`.

openeogeotrellis/opensearch.py
```diff
@@ -22,11 +22,29 @@ def __init__(self, endpoint: str):
...
+                    cache[f["id"]] = f
+                if cache_length_before == len(cache):
```
bossie, line 37: Clarify that this means: if no new items were added to the cache...

Or at least reverse the comparison because this is confusing.

openeogeotrellis/opensearch.py
```diff
@@ -22,11 +22,29 @@ def __init__(self, endpoint: str):
...
+                    # could be because in auto-tests each page contains the same features.
+                    # Or could be because we reached the end of the list.
+                    break
+                if "totalResults" in json and json["totalResults"] <= len(cache):
+                    # This shortcut is not always available in test environment.
```
bossie, line 42: So just remove this check then. Things like `totalResults` have proven to be unreliable in the past so just depend on the actual features returned.

bossie (changes requested):
A couple of remarks inline but my biggest issue is that if you're doing paging, you want to have the results returned in a particular order like here: https://github.com/Open-EO/openeo-opensearch-client/blob/60d668f9fb6c5e191831b345fb72736e0dd87146/src/main/scala/org/openeo/opensearch/backends/OscarsClient.scala#L73

I'm not sure what the particular `sortKeys` value you want in this case but none of "title", "id" and"identifier" seem to work; "start" does but I'm not sure what it means exactly.

openeogeotrellis/opensearch.py
EmileSonneveld: Renamed `features` to `collections`

EmileSonneveld:
With `start` it seems to be sorted on properties>date. Not on the published date, is that stable enough?
image

https://services.terrascope.be/catalogue//collections?startIndex=1&count=200&sortKeys=start
```javascript
data = JSON.parse(document.querySelector("#jsonFormatterRaw > pre").innerText)
console.table(data["features"].map(x=>x["properties"]))
```

openeogeotrellis/opensearch.py
EmileSonneveld: Removed. What a petty. But it should not be too bad on performance.

EmileSonneveld:
I added the sort key anyway. There are multiple products with the same date tough. I guess that after that, it sorted based on when the product was added to the database.

openeogeotrellis/opensearch.py
bossie: I think you renamed the wrong one; so `features` -> `collections` and `f` -> `collection`.

bossie:
> I added the sort key anyway. There are multiple products with the same date tough. I guess that after that, it sorted based on when the product was added to the database.

It's probably me but I don't see how this list is sorted on `date` now.

You could go check with Stijn C. to see if there is a particular `sortKeys` that allows reliable paging or just take the pragmatic approach: remove the paging altogether and increase the page size to e.g. 500.

EmileSonneveld:
Stijn C says that by default, features are sorted by ID: https://git.vito.be/projects/BIGGEO/repos/oscars/browse/src/main/java/be/vito/opensearch/elasticsearch/SearchOperation.java#85
To be sure I added a warning when it misbehaves. (Still running tests)
If in any way, a feature got returned multiple times, the one on the last page would be preserved.
I added a log for this, and it did not happen on CI at least

openeogeotrellis/opensearch.py
EmileSonneveld: fixed

PR #705, opened 2024-03-04
Parse featureflags and tilesize in load_stac.

EmileSonneveld:


EmileSonneveld:
I actually need this feature on `openeo-dev` to continue development. I'll merge it now and we can see for a better solution on monday

PR #724, opened 2024-03-13
More partitions for resample_spatial with higher resolution.

EmileSonneveld:
Increase amount of partitions in proportion to the amount of resolution increased.

### DIFF
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -1237,6 +1237,9 @@ class GeopysparkDataCube(DriverDataCube):
                 estimated_size_in_pixels_tup[0] * resolution_factor[0] / currentTileLayout.tileCols,
                 estimated_size_in_pixels_tup[1] * resolution_factor[1] / currentTileLayout.tileRows
             )
+            # Could also use "currentTileLayout.layoutCols * resolution_factor[0]", but that would be less accurate.
+            # The repartitioning only considers the resolution change. It considers that the partitioning took
+            # already into account band count and the pixel type, sparse/not, ...
             proposed_partition_count = int(proposed_partition_count_tup[0] * proposed_partition_count_tup[1])
             if (  # Only repartition when there would be significantly more
                     max_level.getNumPartitions() * 2 <= proposed_partition_count
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -1241,16 +1241,16 @@ class GeopysparkDataCube(DriverDataCube):
 
             resolution_increase_factor = int(pixel_area_before / pixel_area_after)
             estimated_partitions_tup = (
-                estimated_size_in_pixels_tup[0] * resolution_increase_factor / 256,  # TODO make not hardcoded
-                estimated_size_in_pixels_tup[1] * resolution_increase_factor / 256
+                estimated_size_in_pixels_tup[0] * resolution_increase_factor / 1024,  # TODO make not hardcoded
+                estimated_size_in_pixels_tup[1] * resolution_increase_factor / 1024
             )
             estimated_partitions = int(estimated_partitions_tup[0] * estimated_partitions_tup[1])
             print(estimated_partitions)
-            # if (max_level.getNumPartitions() <= estimated_partitions
-            #         and max_level.layer_type == gps.LayerType.SPACETIME
-            #         and resolution_increase_factor > 2):
-            #     logging.info(f"Repartitioning datacube with {max_level.getNumPartitions()} partitions to {estimated_partitions} before resample_spatial.")
-            #     max_level = max_level.repartition(estimated_partitions)
+            if (max_level.getNumPartitions() <= estimated_partitions
+                    and max_level.layer_type == gps.LayerType.SPACETIME
+                    and resolution_increase_factor > 2):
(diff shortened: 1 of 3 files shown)


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -1222,15 +1222,31 @@ def resample_spatial(
...
-                max_level = max_level.repartition(spatialPartitions*10)
+            resolution_factor = (
+                cellsize_before[0] / resolution[0],
+                cellsize_before[1] / resolution[1]
+            )
+
+            # If the layer had a smaller tileSize specified, there will probably be enough partitions and this logic won't have effect
+            # If the user has a larger tileSize specified, it will probably be max 1024, then we would make more partitions
+            # Anyway, specifying tileSize is an undocumented feature, so we can assume default tileSize
+            default_tile_size = 256
```
EmileSonneveld, line 1239: @jdries Is the reasoning behind this value ok?

jdries: 256 is not necessarily the default, there's also a feature that adapts it dynamically, even if the user does not set it. 
But I'm not sure why we need to guess it, if we can actually get it from the metadata?
I'm wondering if we should include the number of bands, and then perhaps some guesswork on the temporal dimension, to get a guesstimate of memory per partition?

EmileSonneveld: In normal cases, `currentTileLayout.tileCols` could be used, but in `TestMultipleDates.test_resample_spatial` that would be `5`, causing a lot of partitions (slowing the test down x10).

I did not yet find a nice solution that works for normal process graphs and this test. 

For some reason, `layoutCols` is 2 in `test_resample_spatial`, so checking on this value to be 1 to resort to a default won't work

EmileSonneveld: I could use tileCols if I just updated the layer in test_resample_spatial to have more realistic proportions. The test suite runs smooth now, as well as the ERA5 resample test.

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -1242,13 +1254,28 @@ def resample_spatial(
...
         extent = max_level.layer_metadata.layout_definition.extent
         currentTileLayout: gps.TileLayout = max_level.layer_metadata.tile_layout
+        return (
+            (extent.xmax - extent.xmin) / (currentTileLayout.tileCols * currentTileLayout.layoutCols),
+            (extent.ymax - extent.ymin) / (currentTileLayout.tileRows * currentTileLayout.layoutRows)
+        )
 
-        currentResolutionX = (extent.xmax - extent.xmin) / (currentTileLayout.tileCols * currentTileLayout.layoutCols)
-        return currentResolutionX
+    def estimate_layer_size_in_pixels(self):
```
jdries, line 1266: what about taking number of bands into account as well?

tests/test_multiple_date.py
```diff
@@ -300,6 +319,47 @@ def elem(timestamp, value):
 
         return TiledRasterLayer.from_numpy_rdd(LayerType.SPACETIME, rdd, metadata)
 
+    def get_test_layer(self):
```
jdries, line 322: note that there's also some test fixtures somewhere, perhaps those could be improved & reused, or this method could be converted into such a fixture?

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -1222,15 +1222,27 @@ def resample_spatial(
...
-            if(max_level.getNumPartitions() <= spatialPartitions and max_level.layer_type == gps.LayerType.SPACETIME and resolution < 0.5*self.get_cellsize_x()):
-                logging.info(f"Repartitioning datacube with {max_level.getNumPartitions()} partitions to {spatialPartitions*10} before resample_spatial.")
-                max_level = max_level.repartition(spatialPartitions*10)
+            resolution_factor = (
+                cellsize_before[0] / resolution[0],
+                cellsize_before[1] / resolution[1]
+            )
+
+            proposed_partition_count_tup = (
+                estimated_size_in_pixels_tup[0] * resolution_factor[0] / currentTileLayout.tileCols,
```
jdries, line 1237: I'm wondering if this formula is actually a more complex way to write something like:
number_of_spatial_tiles*resolution_factor^2
but anyway, perhaps doesn't matter as I would go for a memory based estimate in the final solution, that we perhaps do after merging this PR?

jdries (changes requested):

openeogeotrellis/geopysparkdatacube.py
EmileSonneveld: It makes a slight difference when layer is not divisible by the tilesize. But is maybe not worth the complexity
In my SPI test it goes from `922` to `1128` when multiplying tile number directly.

tests/test_multiple_date.py
EmileSonneveld: True, moved!

openeogeotrellis/geopysparkdatacube.py
EmileSonneveld: Should not needed here. I added a comment `The repartitioning only considers the resolution change. It considers that the partitioning took already into account band count and the pixel type, sparse/not, ...`

jdries (approved):
Thanks and approved!
Lets merge this one, as it's probably better than the previous guesstimate.
Next step is computing memory per partition in a generic manner.

PR #732, opened 2024-03-19
Allow to load signed result URL in load_ml_model

EmileSonneveld:
Loading from a signed results URL avoids credentials issues.

Another way could have been to sign the content of the `/ml_model_metadata.json` url.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1190,15 +1190,27 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
             with requests.get(model_id) as resp:
                 resp.raise_for_status()
                 metadata = resp.json()
-            if deep_get(metadata, "properties", "ml-model:architecture", default=None) is None:
-                raise OpenEOApiException(
-                    message=f"{model_id} does not specify a model architecture under properties.ml-model:architecture.",
-                    status_code=400)
-            checkpoints = []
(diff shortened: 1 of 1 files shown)


openeogeotrellis/backend.py
```diff
@@ -1190,15 +1190,27 @@ def _set_permissions(job_dir: Path):
...
-            if deep_get(metadata, "properties", "ml-model:architecture", default=None) is None:
-                raise OpenEOApiException(
-                    message=f"{model_id} does not specify a model architecture under properties.ml-model:architecture.",
-                    status_code=400)
-            checkpoints = []
-            assets = metadata.get('assets', {})
-            for asset in assets:
-                if "ml-model:checkpoint" in assets[asset].get('roles', []):
-                    checkpoints.append(assets[asset])
+            if "ml_model_metadata.json" in model_id:
```
JeroenVerstraelen, line 1193: Maybe use endswith instead of in. I don't think this edge case will ever occur but better to be safe.

JeroenVerstraelen:
The code looks good but I'm not sure why ml_model_metadata.json has a different json structure compared to other files. I suspect the signed urls are the ones that don't end with 'ml_model_metadata.json'?

EmileSonneveld:
PR put on draft and might close later.
I'll implement the other way, signing the urls in `/ml_model_metadata.json` when the request self is signed...

PR #738, opened 2024-03-25
Special temporal step for COPERNICUS_30.

EmileSonneveld:
@soxofaan , I tested this against a whole bunch of process graphs. Can I commit them in `openeo-geopyspark-integrationtests`?

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1635,6 +1635,9 @@ class GpsProcessing(ConcreteProcessing):
                     nr_bands = len(band_names)
 
                     temporal_step = metadata.get("cube:dimensions", "t", "step", default=None)
+                    if collection_id == 'COPERNICUS_30':
+                        logger.info(f"Special temporal step for {collection_id}")
+                        temporal_step = "P9999Y"
 
                     if collection_id == 'TestCollection-LonLat4x4':
                         # This layer is always 4x4 pixels, adapt resolution accordingly


openeogeotrellis/backend.py
```diff
@@ -1635,6 +1635,9 @@ def extra_validation(
                     nr_bands = len(band_names)
 
                     temporal_step = metadata.get("cube:dimensions", "t", "step", default=None)
+                    if collection_id == 'COPERNICUS_30':
```
soxofaan, line 1638: I don't think openeo-geopyspark-driver should refer to hardcoded real collections  this way. 
openeo-geopyspark-driver is essentially just a reusable library and the actual collection definition is in other projects (like openeo-deploy).

In this case, isn't it possible to detect this condition from loaded "layercatalog" data?

EmileSonneveld: It would be possible to add `"t": {... "step": "P10000D"}` in the `layercatalog`. 
Otherwise, it is possible to add a temporal extent to the layer in catalog, but there is something special going on with the dates of the tiles I don't quite understand

soxofaan: Instead of these dummy steps, isn't it easier to specify in the layer catalog that there is no actual time dimension, or that there is just one label in the time dimension?

EmileSonneveld: That might be too drastic. The time dimension is there. Tiles have different dates, but there is just one tile per spatial key

soxofaan: If there are different dates, then the metadata should not use a dummy step of ["10000 years"](https://imgflip.com/i/8mxlye) either. 

In case of COPERNICUS_30 I think there are just a handful of dates, so I would use the "values" field of the [temporal dimension](https://github.com/stac-extensions/datacube?tab=readme-ov-file#temporal-dimension-object) properly, instead of abusing the "step" field. Which we can use then to properly estimate the number of expected observations in a given time window

EmileSonneveld: True, I queried the files and only got those days out of it. Just using the first and the last as a temporal extent does the trick:
```
'2019-10-25'
'2019-10-26'
'2019-10-27'
'2019-10-28'
'2019-10-30'
'2019-10-31'
'2019-11-01'
'2019-11-02'
'2019-11-03'
'2019-11-04'
'2019-11-05'
```
WIll commit later

soxofaan:
is it ok to remove this feature branch? To avoid stale branches polluting jenkins listings

EmileSonneveld:
sure, removed

PR #741, opened 2024-03-26
Raise clear error for 'SENTINEL_5P_L2' with only 'dataMask'

EmileSonneveld:
There was already a check on `collection_id` here, so a test for `SENTINEL_5P_L2` seems acceptable. 
If more checks like this get added in the future something more generic in the layercatalog might be nice.

### DIFF
--- a/openeogeotrellis/layercatalog.py
+++ b/openeogeotrellis/layercatalog.py
@@ -471,8 +471,15 @@ class GeoPySparkLayerCatalog(CollectionCatalog):
                 return (pyramid_factory.datacube_seq(projected_polygons_native_crs, from_date, to_date,metadata_properties(),collection_id,datacubeParams) if single_level
                         else pyramid_factory.pyramid_seq(extent, srs, from_date, to_date))
             else:
-                if collection_id == 'PLANETSCOPE':
+                shub_band_names = metadata.band_names
 
+                if collection_id == 'SENTINEL_5P_L2':
+                    if shub_band_names == ["dataMask"]:
+                        raise OpenEOApiException(
+                            f"Cant load collection '{collection_id}' with only 'dataMask' band. Add 1 other band to make it work.",
+                            status_code=400)
+
+                if collection_id == 'PLANETSCOPE':
                     if 'byoc_collection_id' in feature_flags:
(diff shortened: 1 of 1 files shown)


jdries:
seems to be a spelling error 'cant' in the error message
should we immediately add a second error for the case where multiple bands are used? (Which also won't work afaik.)

EmileSonneveld:
When using multiple bands, the error from SH is already clear:
`"The script can only use one product type at once! Instead it uses: [CO, O3]."`

Typo fixed

EmileSonneveld:
The error when using multiple bands is actually quite verbose and take a while. I'll block it early too.
`OpenEO batch job failed: Exception during Spark execution: org.openeo.geotrellissentinelhub.SentinelHubException: Sentinel Hub returned an error response: HTTP/1.1 400 Bad Request with body: {"error":{"status":400,"reason":"Bad Request","message":"The script can only use one product type at once! Instead it uses: [O3, CO].","code":"RENDERER_EXCEPTION"}} request: POST https://creodias.sentinel-hub.com/api/v1/process with (possibly abbreviated) body: { "input": { "bounds": { "bbox": [5.540042760864049, 42.90896435887362, 19.5082967291157, 51.79785324776091], "properties": { "crs": "http://www.opengis.net/def/crs/EPSG/0/4326" } }, "data": [ { "type": "sentinel-5p-l2", "dataFilter": {"timeRange":{"from":"2023-04-11T00:00:00Z","to":"2023-04-12T00:00:00Z"}}, "processing": {} } ] }, "output": { "width": 256, "height": 256, "responses": [ { "identifier": "default", "format": { "type": "image/tiff" } } ] }, "evalscript": "//VERSION=3\nfunction setup() {\n return {\n input: [{\n \"bands\": [\"CO\", \"O3\"]\n }],\n output: {\n bands: 2,\n sampleType: \"FLOAT32\",\n }\n };\n}\n\nfunction evaluatePixel(sample) {\n return [sample.CO, sample.O3];\n}" }
`

EmileSonneveld:
@jdries Good message?` "Collection 'SENTINEL_5P_L2' got requested with multiple bands: ['CO', 'O3']. Only one band is supported, with or without the 'dataMask' band."`

jdries:
Sure!

PR #757, opened 2024-04-18
Better temporal intersection for extra_validation

EmileSonneveld:
Committed script to find temporal extent for layer here: https://git.vito.be/projects/BIGGEO/repos/openeo-deploy/pull-requests/15/overview

First the code would only take the temporal extent from the layer catalog when the user did not specify something. But for one process graph a user specified a huge and wrong temporal extent, confusing the `is_layer_too_large` code. Now the code takes an intersection of the user and catalog temporal extent.


EmileSonneveld:
Wait, I did a huge `COPERNICUS_30` query and got different dates than what I extracted from the data.
For North East Africa I got dates from `2012-05-05` to `2014-12-01`

EmileSonneveld:
@soxofaan would this in the layer catalog be fine?
`"temporal": {"interval": [["2010-12-12", null]], "step": "infinite"}` (non standard)
or 
`"temporal": {"interval": [["2010-12-12", null]], "skip_validation": true}`
or maybe add `skip_validation` to some other part of the layer

openeogeotrellis/backend.py
```diff
@@ -1153,6 +1156,98 @@ def request_costs(
 
 
 class GpsProcessing(ConcreteProcessing):
+
+    @staticmethod
+    def derive_temporal_extent(collection_id: str, constraints: dict, env: EvalEnv) -> List[Optional[str]]:
```
soxofaan, line 1161: Why is this under GpsProcessing? Isn't this more catalog aspect, so it better fits under GeoPySparkLayerCatalog , or even  a toplevel function at openeogeotrellis.layercatalog?

openeogeotrellis/backend.py
```diff
@@ -1153,6 +1156,98 @@ def request_costs(
...
+
+            temporal_extent = (
+                max(beginnings),  # ISO date is sortable like a string
+                min(ends),
+            )
+        return temporal_extent
+
+    @staticmethod
+    def estimate_number_of_temporal_observations(collection_id: str,
+                                                 constraints: dict,
```
soxofaan, line 1210: Same question if this conceptually belongs under GpsProcessing

soxofaan (commented):
didn't go deep yet, just some quick notes

openeogeotrellis/backend.py
EmileSonneveld: Fits nice there. Moved!

openeogeotrellis/backend.py
EmileSonneveld: Also moved to GeoPySparkLayerCatalog

openeogeotrellis/layercatalog.py
```diff
@@ -800,6 +800,93 @@ def _native_crs(self, metadata: GeopysparkCubeMetadata) -> str:
...
+        # https://github.com/stac-extensions/datacube?tab=readme-ov-file#temporal-dimension-object
+        temporal_step = parse_approximate_isoduration(temporal_step)
+        estimate_days_per_sample = temporal_step.days
+
+        from_date, to_date = temporal_extent
+        if to_date is None:
+            if from_date is None:
+                days = 1
+                logger.warning(
+                    f"Got open temporal extent: {repr(temporal_extent)}. Assuming {days} day."
```
soxofaan, line 875: Python pro-tip: use the [`=` specifier in f-strings](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging) to simplify these kind of logging constructs.

for example:
```suggestion
                    f"Got open {temporal_extent=}. Assuming {days=}."
```

openeogeotrellis/layercatalog.py
```diff
@@ -800,6 +800,93 @@ def _native_crs(self, metadata: GeopysparkCubeMetadata) -> str:
...
+                logger.warning(
+                    f"Got half open temporal extent: {repr(temporal_extent)}. Assuming it goes till today."
+                )
+                from_date_parsed = dateutil.parser.parse(from_date).replace(tzinfo=None)
+                to_date_now = datetime.now().replace(tzinfo=None)
+                days = (to_date_now - from_date_parsed).days / estimate_days_per_sample
+        else:
+            days = (dateutil.parser.parse(to_date) - dateutil.parser.parse(
+                from_date)).days / estimate_days_per_sample
+        days = max(int(days), 1)
```
soxofaan, line 887: did I miss something or is the cast  `int(days)` redundant?

openeogeotrellis/layercatalog.py
```diff
@@ -1134,8 +1221,7 @@ def check_missing_products(
 def is_layer_too_large(
         spatial_extent: dict,
         geometries: Union[DriverVectorCube, DelayedVector, BaseGeometry],
-        temporal_extent: Tuple[str, str],
-        temporal_step:str,
+        number_of_temporal_observations: float,
```
soxofaan, line 1224: shouldn't the type hint of a number be `int`?

openeogeotrellis/layercatalog.py
```diff
@@ -800,6 +800,93 @@ def _native_crs(self, metadata: GeopysparkCubeMetadata) -> str:
...
+        from_date, to_date = temporal_extent
+        if to_date is None:
+            if from_date is None:
+                days = 1
+                logger.warning(
+                    f"Got open temporal extent: {repr(temporal_extent)}. Assuming {days} day."
+                )
+            else:
+                logger.warning(
+                    f"Got half open temporal extent: {repr(temporal_extent)}. Assuming it goes till today."
```
soxofaan, line 879: same:
```suggestion
                    f"Got half open {temporal_extent=}. Assuming it goes till today."
```

soxofaan (commented):
some minor notes, but nothing really blocking

soxofaan (approved):

PR #759, opened 2024-04-22
Change max_executor_or_driver_memory to 16. 64 was a very high limit

EmileSonneveld:


tcassaert:
Is this 16GB a combination of overhead and "regular" memory?

EmileSonneveld:
@tcassaert It goes through tests. Is the limit fine, or too low? I often specified 20Gb `driver-memory` in the past, so I got the feeling there is space for more

jdries:
Pay attention that things depend on the environment. Terrascope has higher limits, cdse limit is 15GB....
________________________________

tcassaert:
@EmileSonneveld for CDSE the limit the user sees should then be 15GB instead of 16GB.

jdries:
Config adjusted in the right place, closing this one.

PR #765, opened 2024-04-25
Truncate udf error in middle

EmileSonneveld:


openeogeotrellis/backend.py
```diff
@@ -1017,6 +1017,13 @@ def get_missing_sentinel1_band() -> Optional[str]:
...
                     udf_stacktrace = GeoPySparkBackendImplementation.extract_udf_stacktrace(root_cause_message)
                     if udf_stacktrace:
+                        if len(udf_stacktrace) > width - 100:
+                            udf_stacktrace_list = udf_stacktrace.split("\n")
+                            udf_stacktrace_new = "\n".join(
+                                udf_stacktrace_list[:5] + ["... skipped stack frames ..."] + udf_stacktrace_list[-5:]
+                            )
+                            if len(udf_stacktrace_new) < width - 100:
+                                udf_stacktrace = udf_stacktrace_new
                         summary = f"UDF Exception during Spark execution: {udf_stacktrace}"
```
jdries, line 1027: can we additionally not mention Spark here? That's an implementation detail which has little meaning for the user.
Better:
UDF exception while evaluating processing graph. Please check your user defined functions.

jdries (approved):
added minor comment which is not about the actually changed code
so approving already

openeogeotrellis/backend.py
EmileSonneveld: updated!

Issue #460, opened 2023-06-22
integrate CLMS HRL VPP


jdries:
https://land.copernicus.eu/pan-european/biophysical-parameters/high-resolution-vegetation-phenology-and-productivity
problem here is that there's one collection, with multiple producttypes, and each type has a single band
this should become one collection with multiple bands

jdries:
@JohanKJSchreurs  This is a good candidate to test our new STAC api.
This collection already exists in opensearch catalog as well, mostly a matter of porting metadata to STAC.

jdries:
Collection id: copernicus_r_3035_x_m_hrvpp-vpp_p_2017-now_v01_openeo

JohanKJSchreurs:
We are implementing this in the [stac-catalog-builder project](https://github.com/VitoTAP/stac-catalog-builder).

The issue linked below is the main one that contains a breakdown of the parts/features we need for this integration
https://github.com/VitoTAP/stac-catalog-builder/issues/16

JohanKJSchreurs:
The implementation in the [stac-catalog-builder](https://github.com/VitoTAP/stac-catalog-builder) is complete: https://github.com/VitoTAP/stac-catalog-builder/issues/16

Some small improvements can still be done, to update the collections & items in the STAC API with improved information, but those should be separate GH issues.

At present, four of the VPP collections have been converted and uploaded now to the development environment of the terra-stac-api at VITO.
The largest collection has many products (6.5 million) and we will need to download that in several pieces because the process runs too long.

JohanKJSchreurs:
How we can solve the long download of the large collection:

We can add some options to the command to tell it what time slice to download.
Right now it tries to download the entire collection, that is tot say the entire period.

We already divide that entire period into smaller time slots in order to limit the number of products in each query to a reasonable number. So if we add options for a start and end date then we could do a partial download. 
That way we can test just upload and test the collection with a more limited set of STAC items

Furthermore, with same additional work we could download the whole collection in several parts, and in each run save out the STAC items for just those slices.
We can already upload upload the STAC items in several parts or sets.
At present, the collection.json file would be created and overwritten every time on disk when you do a partial download, but the collection file would still have the same data anyway. (It does not link to its STAC items in this case because there are far too many items for a static STAC collection).  However with a little extra work we could split up the command so one command downloads/creates the "empty" collection and another command downloads/creates the STAC items.

VictorVerhaert:
Update:
[...]

PR #767, opened 2024-04-29
add tests for property filters

bossie:
Part of #460.


bossie:
@soxofaan,

I'm trying to address this TODO here but can't seem to find the way to implement it: https://github.com/Open-EO/openeo-geopyspark-driver/blob/66da61602f36ff99df2d24c56bd8690bf40fd81b/tests/test_api_result.py#L4006-L4007

What the test does is pass a property filter (`"season" == "s1"`) to `load_stac` so only STAC Item "item01" is loaded ("item02" doesn't have this "season" property).

To this end, I'm trying to see if (in pseudocode) `org.openeo.geotrellis.file.FixedFeaturesOpenSearchClient().addFeature()` is called exactly once with `Feature("item01")`: https://github.com/Open-EO/openeo-geopyspark-driver/blob/66da61602f36ff99df2d24c56bd8690bf40fd81b/openeogeotrellis/load_stac.py#L324

Some guidance?

PR #801, opened 2024-06-10
use DriverVectorCube in chunk_polygon Open-EO/openeo-python-driver#288

### LINKED ISSUE #288: apply_polygon (previously known as chunk_polygon) abuses MultiPolygon as collection of polygons in it's internal API
Stumbled on this while working on #287:

The `apply_polygon` process implementation (originally added as `chunk_polygon` in https://github.com/Open-EO/openeo-python-driver/commit/a411109cbf909a316bac5ea2945c20b4a27c932d)  coerces the provided geometries into a MultiPolygon and pushes this API through into geopyspark driver, e.g.
https://github.com/Open-EO/openeo-geopyspark-driver/blob/1f22c0da34535e986908480673ad341c9f89324d/openeogeotrellis/geopysparkdatacube.py#L572-L575

MultiPolygon should not be abused as a collection of polygons. Instead the API should use something like a FeatureCollection or vector cube abstraction

JeroenVerstraelen:


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -572,9 +572,7 @@ def partitionByKey(spatialkey):
     def chunk_polygon(
         self,
         reducer: Union[ProcessGraphVisitor, Dict],
-        # TODO: it's wrong to use MultiPolygon as a collection of polygons. MultiPolygons should be handled as single, atomic "features"
-        #       also see https://github.com/Open-EO/openeo-python-driver/issues/288
-        chunks: MultiPolygon,
+        chunks: DriverVectorCube,
```
soxofaan, line 575: This changes an API signature defined in openeo-python-driver. I don't think there is PR for that already

JeroenVerstraelen: There is, you can find it here: https://github.com/Open-EO/openeo-python-driver/pull/291

PR #830, opened 2024-07-19
get_best_url() giving priority to local path if available. 

EmileSonneveld:


openeogeotrellis/load_stac.py
```diff
@@ -505,6 +506,25 @@ def intersects_temporally(interval) -> bool:
...
+    alternate = asset.extra_fields.get("alternate")
+    if alternate:
+        for key, alternate_local in alternate.items():
+            if key != "local":
+                continue
+            href = alternate_local.get("href")
+            # Checking if file exists takes around 10ms on /data/MTDA mounted on laptop
+            # Checking if URL exists takes around 100ms on https://services.terrascope.be
+            # Checking if URL exists depends also on what Datasource is used in the scala code.
+            # That would be hacky to predict here.
```
EmileSonneveld, line 522: @VictorVerhaert Is this logic fine for you?

openeogeotrellis/load_stac.py
```diff
@@ -505,6 +506,25 @@ def intersects_temporally(interval) -> bool:
...
+            if Path(urlparse(href).path).exists():
```
soxofaan, line 523: This looks pretty weird: you just check if the path of a href exists locally.
Can you give an example where that would make sense?

EmileSonneveld: It only checks when for a "local" alternate url. Example data:
```json
{
    "B02_10m": {
        "href": "https://services.terrascope.be/download/CGS_S2_L2A/2024/07/18/S2A_MSIL2A_20240718T123311_N0510_R052_T28VDR_20240718T165647/S2A_MSIL2A_20240718T123311_N0510_R052_T28VDR_20240718T165647.SAFE/GRANULE/L2A_T28VDR_A047382_20240718T123307/IMG_DATA/R10m/T28VDR_20240718T123311_B02_10m.jp2",
        "type": "image/jp2",
        // ...
        "alternate": {
            "local": {
                "href": "file:///data/MTDA/CGS_S2/CGS_S2_L2A/2024/07/18/S2A_MSIL2A_20240718T123311_N0510_R052_T28VDR_20240718T165647/S2A_MSIL2A_20240718T123311_N0510_R052_T28VDR_20240718T165647.SAFE/GRANULE/L2A_T28VDR_A047382_20240718T123307/IMG_DATA/R10m/T28VDR_20240718T123311_B02_10m.jp2"
            }
        },
        // ...
    }
}
```
https://stac.terrascope.be/collections/sentinel-2-l2a/items/S2A_MSIL2A_20240718T123311_N0510_R052_T28VDR_20240718T165647

openeogeotrellis/load_stac.py
VictorVerhaert: In some collections, the alternate link is not local but 's3', in those cases when we have access to this s3 it should be used.
Note that these s3 alternate links might contain extra fields indicating an endpoint or region.

I might suggest using this package instead of path: https://pypi.org/project/universal-pathlib/ in those cases, which allows the usage of both paths as well as s3 links.

openeogeotrellis/load_stac.py
soxofaan: I would then also include a check that the protocol is `file://`

EmileSonneveld: Added. 
Btw, not sure if it is cost effective to add an auto test for this. Breaking `def load_stac` in pieces will be a big effort, and adding a semi-external stac catalog to integration tests might break in the future.

openeogeotrellis/load_stac.py
EmileSonneveld: S3 would be a different code path. To support it, it would be nice to have an example stac catalog that uses this feature.
The `local` path can refer to a mounted s3 bucket tough, like `/eodata`.

openeogeotrellis/load_stac.py
soxofaan: you can just unit-test  `get_best_url` directly, right? Should be pretty simple (no need for complex mocking), and acts as added context/docs what this feature is about.

openeogeotrellis/load_stac.py
```diff
@@ -505,6 +506,30 @@ def intersects_temporally(interval) -> bool:
     return GeopysparkDataCube(pyramid=gps.Pyramid(levels), metadata=metadata)
 
 
+def get_best_url(asset):
```
soxofaan, line 509: maybe add a typehint for `asset` as well

openeogeotrellis/load_stac.py
EmileSonneveld: It works for mounted s3 buckets now. But I can not test it on `/eodata/Global-Mosaics/` specifically yet. That can be done, when the `eo:bands` property is supported, or when Cloudferro's gives access: `CDSE-1342`

openeogeotrellis/load_stac.py
EmileSonneveld: Creating the stac asset directly did not initialize the alternate properties. Added a test now that loads a small region of a mocked catalog.

openeogeotrellis/load_stac.py
```diff
@@ -229,8 +230,8 @@ def operator_value(criterion: Dict[str, object]) -> (str, object):
             else:
                 modifier = None
                 # standard behavior seems to be to include only a minimal subset e.g. https://stac.openeo.vito.be/
-                fields = [f"properties.{property_name}" for property_name in
-                          {"proj:epsg", "proj:bbox", "proj:shape"}.union(all_properties.keys())]
+                fields = list(sorted([f"properties.{property_name}" for property_name in
+                          {"proj:epsg", "proj:bbox", "proj:shape"}.union(all_properties.keys())]))
```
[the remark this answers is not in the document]
EmileSonneveld: top, used! Also enabled `pre-commit`. It won't help for this case specific, but hopefully resolve other issues

tests/test_api_result.py
```diff
@@ -1927,6 +1928,24 @@ def urllib_mock() -> UrllibMocker:
         yield mocker
 
 
+class UrllibAndRequestMocker:
```
EmileSonneveld, line 1931: @bossie Moved the double mock to here. It is almost a drop-in-replace for the original mocks. Ok?

bossie (approved):

PR #836, opened 2024-08-19
Use json_default 

EmileSonneveld:
https://github.com/Open-EO/openeo-python-driver/issues/303


EmileSonneveld:
@jdries This MR is better than the one from last week.

PR #843, opened 2024-08-21
Read separate_asset_per_band and pass it to Scala code.

EmileSonneveld:
(Depends on openeo-geotrellis-extensions code)


jdries (approved):

EmileSonneveld:
The tests seems to be stuck on artifactory at the moment. It would be nice to run them again with the new geotrellis extensions before merging

PR #848, opened 2024-08-27
Batch job administration cost

bossie:
https://github.com/eu-cdse/openeo-cdse-infra/issues/244


openeogeotrellis/config/config.py
```diff
@@ -131,6 +131,7 @@ class GpsBackendConfig(OpenEoBackendConfig):
     default_usage_cpu_seconds: float = 1.5 * 3600
     default_usage_byte_seconds: float = 3 * 1024 * 1024 * 1024 * 3600
     report_usage_sentinelhub_pus: bool = True
+    batch_job_administration_cost_credits: Optional[float] = None
```
soxofaan, line 134: FYI: when discussing this feature I used "administration cost" just as a matter of speaking. Maybe it's more generic and simpler to just call this a "base fee"

PR #851, opened 2024-08-28
Check memory size after taking into account python_max.

EmileSonneveld:
Moved python_max calculations higher up in the function where test code can still reach.
Throw error when asking for too much memory.

_EDIT: No changes to memory calculation_


EmileSonneveld:
@soxofaan , looks fine? That `memOverheadBytes` was not taken into account was curious for me. We can also wait till Jeroen Dries is back

EmileSonneveld:
This passed CI, but the github UI is not updated yet

PR #872, opened 2024-09-17
support EJR in async_task

bossie:
https://github.com/eu-cdse/openeo-cdse-infra/issues/261


bossie:
- Reverts SHub sync request workaround (introduced because of `async_task`'s dependency on `ZkJobRegistry`);
- enables EJR in `async_task`.

Issue #812, opened 2024-06-24
load_stac: align DataCubeParameters with load_collection


jdries:
load_stac has a very minimal setup of DataCubeParameters:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/7aee61514f7f1ab07bbe5ef48033ff056b07cfbf/openeogeotrellis/load_stac.py#L449

It should take over the implementation from load_collection.

Important things to set for sure are target crs, resolution, buffer size. Most of this can be had from LoadParameters.

PR #875, opened 2024-09-20
align DataCubeParameters with load_collection

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/812


bossie:
`load_stac` now reuses dedicated method to create `DataCubeParameters` from `LoadParameters`. It's in a new module because otherwise there's a circular reference (`layercatalog.py` depends on `load_stac.py`).

This dedicated method makes use of some constants (in this case: `EvalEnv` keys) so to avoid another circular reference (`layercatalog.py` depends on `datacube_parameters.py`), these constants have been put into another module: `constants.py`.

bossie:
Ok to put `EvalEnv` keys:

- in a class called `EvalEnvKeys`;
- in a module called `constants.py`?

or is there a more Pythonic way that you prefer @soxofaan ?

bossie:
I aligned it with some other containers-of-constants in the code like `YARN_STATE`, `JOB_STATUS` etc.

PR #882, opened 2024-09-25
add support for apply_vectorcube udf #856

JeroenVerstraelen:


JeroenVerstraelen:
Related PR in python-client:
https://github.com/Open-EO/openeo-python-client/pull/631

JeroenVerstraelen:
Related PR in openeo-python-driver:
https://github.com/Open-EO/openeo-python-driver/pull/314/files

soxofaan:
after reviewing the other PRs in python client and python driver:  do we actually need this #882 that adds a test in geopyspark driver actually? All functionality is actually elsewhere: python client and python driver, so the tests should be there

JeroenVerstraelen:
> after reviewing the other PRs in python client and python driver: do we actually need this #882 that adds a test in geopyspark driver actually? All functionality is actually elsewhere: python client and python driver, so the tests should be there

The test in the geopyspark driver is specifically to see if you can apply an udf on the output of raster_to_vector without any issues. Because previously that didn't work as expected.

soxofaan:
ah ok indeed, `raster_to_vector` is the differentiating aspect here

soxofaan:
Also to include in this PR: as it depends on python client and python driver changes, it should bump the version constraint of these dependencies in setup.py

PR #901, opened 2024-10-09
Rasample zero

EmileSonneveld:


jdries (approved):
As far as I can tell, the modified code is limited to the actual 'layer too large' check, hence fine for me!

EmileSonneveld:
The logic is also quite lookalike to here: https://github.com/Open-EO/openeo-geopyspark-driver/blob/0200f28f846c6b465cf5372b67d8671ce36f8d9b/openeogeotrellis/layercatalog.py#L244

Issue #845, opened 2024-08-22
use zip files for UDF dependencies


JeroenVerstraelen:
A more permanent solution for https://github.com/eu-cdse/openeo-cdse-infra/issues/112

JeroenVerstraelen:
To be checked:
- Does the windmill application work with the current solution? (No zip)
- Does fsync in python solve the initial problem?

JeroenVerstraelen:
Blocked by fusemount

soxofaan:
I decided to merge #914 .  I tried to cover as much as possible with unit testing, but some parts where not feasible (also see #915). I had hoped to also do some real testing on k8s, but it was unclear how much additional time I would need for that (e.g. setting up this skaffold thing). This PR also just adds a new feature as opt-in and should not change the current behavior.

soxofaan:
enabled ZIP based UDF dependency handling  on CDSE dev with
https://git.vito.be/projects/TPT/repos/os_creodias_openeo_k8s/commits/657d38743ecacf73efdc40a5519cf3faa97d615c#kube_resources%2Fapplications%2Fopeneo_config%2Fbackendconfig_dev.py

soxofaan:
I hope with CDSE integration tests now pass with these latest commits https://github.com/Open-EO/openeo-geopyspark-driver/commit/50a127a438dfa8327a54e0bdc4e52adc782bb51e and  https://github.com/Open-EO/openeo-geopyspark-driver/commit/84aecb7ca2bd7a751a7e0f605257974b209effab

soxofaan:
CDSE integration tests now passed

relevant logs from the related "test_udf_dependency_handling" job j-241030b14656437fa80e80da41ff9600

driver:
```
Installing Python UDF dependencies with ['/opt/venv/bin/python3', '-m', 'pip', 'install', '--target', '/opt/spark/work-dir/tmp/udfpydeps-pack-zvnufi85/packages', '--progress-bar', 'off', '--disable-pip-version-check', '--no-input', '--retries', '2', '--timeout', '20', 'alabaster==0.7.13']: start 2024-10-30 11:13:51.678563
pip install exited with exit code 0
[...]

PR #914, opened 2024-10-22
zip archive based UDF dependency handling

soxofaan:
self-review PR for #845


soxofaan:
- [ ] add changelog entry and bump version?

PR #916, opened 2024-10-23
Wait till path available

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/329

### DIFF
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -356,11 +356,10 @@ def run_job(
                 ml_model_metadata = result.get_model_metadata(str(output_file))
                 logger.info("Extracted ml model metadata from %s" % output_file)
             for name, asset in the_assets_metadata.items():
-                # TODO: test in separate branch
-                # if not asset.get("href").lower().startswith("s3:/"):
-                #     # fusemount could have some delay to make files accessible, so poll a bit:
-                #     asset_path = get_abs_path_of_asset(asset["href"], job_dir)
-                #     wait_till_path_available(asset_path)
+                if not asset.get("href").lower().startswith("s3:/"):
+                    # fusemount could have some delay to make files accessible, so poll a bit:
+                    asset_path = get_abs_path_of_asset(asset["href"], job_dir)
+                    wait_till_path_available(asset_path)
                 add_permissions(Path(asset["href"]), stat.S_IWGRP)
             logger.info(f"wrote {len(the_assets_metadata)} assets to {output_file}")
             assets_metadata.append(the_assets_metadata)
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -356,9 +356,12 @@ def run_job(
                 ml_model_metadata = result.get_model_metadata(str(output_file))
                 logger.info("Extracted ml model metadata from %s" % output_file)
             for name, asset in the_assets_metadata.items():
-                if not asset.get("href").lower().startswith("s3:/"):
+                href = asset["href"]
+                url = urlparse(href)
+                if url.scheme in ["", "file"]:
+                    file_path = url.path
                     # fusemount could have some delay to make files accessible, so poll a bit:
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -179,6 +182,36 @@ def _setup_local_spark(out: TerminalReporter, verbosity=0):
 def api_version(request):
     return request.param
 
+# TODO: Deduplicate code with openeo-python-driver
```
EmileSonneveld, line 185: @soxofaan Is there an easy way to share test helpers? It is not very common to make test code available from other repositories as far as i know

soxofaan: that _Sleeper implementation actually comes from python client I think. There I started putting testing utilities  under the `openeo.testing` namespace: https://github.com/Open-EO/openeo-python-client/tree/master/openeo/testing

We also have that with python driver: https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/testing.py

openeogeotrellis/deploy/batch_job.py
```diff
@@ -356,11 +356,10 @@ def run_job(
                 ml_model_metadata = result.get_model_metadata(str(output_file))
                 logger.info("Extracted ml model metadata from %s" % output_file)
             for name, asset in the_assets_metadata.items():
-                # TODO: test in separate branch
-                # if not asset.get("href").lower().startswith("s3:/"):
-                #     # fusemount could have some delay to make files accessible, so poll a bit:
-                #     asset_path = get_abs_path_of_asset(asset["href"], job_dir)
-                #     wait_till_path_available(asset_path)
+                if not asset.get("href").lower().startswith("s3:/"):
```
soxofaan: `asset.get("href")` could return None which will fail hard, so I would harden that with:

```suggestion
                if not asset.get("href", "").lower().startswith("s3:/"):
```

tests/conftest.py
EmileSonneveld: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/329#issuecomment-2432635380

openeogeotrellis/deploy/batch_job.py
soxofaan: and instead of skipping this arbitrary when `startswith("s3:/")`,
isn't it more robust to only do it the "href" looks like a local path: e.g. it starts with a slash  or a `file:/` protocol.

openeogeotrellis/deploy/batch_job.py
```diff
@@ -356,11 +356,13 @@ def run_job(
...
-                #     # fusemount could have some delay to make files accessible, so poll a bit:
-                #     asset_path = get_abs_path_of_asset(asset["href"], job_dir)
-                #     wait_till_path_available(asset_path)
+                href = asset["href"]
+                url = urlparse(href)
+                if url.scheme in ["", "file"]:
+                    file_path = url.path
+                    # fusemount could have some delay to make files accessible, so poll a bit:
+                    asset_path = get_abs_path_of_asset(file_path, job_dir)
+                    wait_till_path_available(asset_path)
```
soxofaan, line 365: about `wait_till_path_available` : I think it's overkill to implement exponential back-off here as the "exist" check is pretty cheap.
Now with the exponential backoff, in the last try there is a sleep of 2 minutes (128s), which is very long. I would just try every 5 seconds up to 1 or 2 minutes in total

EmileSonneveld: The exists check could make a request to S3 if it is a mount. So multiple jobs doing a query at the same rate might impact the S3 connection.
Only in the test suite it refers to local files I think

soxofaan: true, an "exist" over S3 is probably not that cheep,  
but still, doing exponential sleep with base 2 (8s, 16s, 32s, 64s and 128s at the moment)  feels a bit pessimistic

EmileSonneveld: Changed! I am merging already to test on staging again

PR #918, opened 2024-10-24
Explicitly set all stac fields

EmileSonneveld:
Tested on `hrvpp-vi` with http://stac.openeo.vito.be and http://stac-openeo-dev.vgt.vito.be
(Automatic test needs to be enabled after merge: https://git.vito.be/projects/TPT/repos/os_creodias_openeo_k8s/commits/1bf7dbf0d2d16e17ef068ae5b1f0fc613f92b470 )

Tested on https://stac.terrascope.be/collections/sentinel-2-l2a

And of course the standard test suite

### DIFF
--- a/openeogeotrellis/load_stac.py
+++ b/openeogeotrellis/load_stac.py
@@ -239,6 +239,7 @@ def load_stac(url: str, load_params: LoadParameters, env: EvalEnv, layer_propert
             else:
                 modifier = None
                 # standard behavior seems to be to include only a minimal subset e.g. https://stac.openeo.vito.be/
+                # and e.g. https://stac.terrascope.be
                 # include all fields to ease the transition of stac catalog
                 fields = [
                     "type",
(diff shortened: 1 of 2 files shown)


openeogeotrellis/load_stac.py
```diff
@@ -239,10 +239,18 @@ def operator_value(criterion: Dict[str, object]) -> (str, object):
             else:
                 modifier = None
                 # standard behavior seems to be to include only a minimal subset e.g. https://stac.openeo.vito.be/
-                fields = sorted(
-                    f"properties.{property_name}"
-                    for property_name in {"proj:epsg", "proj:bbox", "proj:shape"}.union(all_properties.keys())
-                )
+                # include all fields to ease the transition of stac catalog
```
EmileSonneveld, line 242: @bossie , would it make sense to only do this when requesting from https://stac.openeo.vito.be ?

jdries:
@EmileSonneveld why don't we just push on with the update of stac.openeo.vito.be? From what I understood, stac.terrascope.be has already been upgraded, so we're anyway forced to support the new behaviour already?

EmileSonneveld:
stac.terrascope.be seems updated already indeed. So https://stac.terrascope.be/collections/sentinel-2-l2a is broken until this is merged.
The code proposed here works with the new and the old VITO stac

Issue #920, opened 2024-10-25
resample_spatial of SENTINEL3_SYN_L2_SYN


bossie:
https://forum.dataspace.copernicus.eu/t/spatial-resampling-causes-arraymemoryerror-and-catalogue-validation-error/1288/3

bossie:
Fixed by returning a data cube in EPSG:4326 and thereby provoking an actual call to `resample_spatial` (+ some fixes related to the hourly resolution of the Sentinel 3 collections).

PR #922, opened 2024-10-29
enforce actual call to resample_spatial

bossie:
#920


bossie:
Branched from wrong... branch.

Issue #883, opened 2024-09-26
export_workspace: skip writing to local storage


jdries:
When doing save_result => export_workspace,
the data is stored once in default location and once in workspace.
This is problematic for huge processing jobs, because we will have to pay for the storage.

bossie:
@jdries What are the implications from the user's point of view? Does this mean that the STAC Collection of batch job results has no items then (also because OpenEO won't be able to serve the assets anymore)?

jdries:
or can we make those items point to the location where we exported the results to? 
If complex, the main thing to achieve here is simply that the file is gone, potentially triggered by a feature flag or job option if the impact is too unpredictable.

bossie:
Currently this job_option will remove the exported assets from the default workspace (the batch job output directory):

```json
{
  "remove-exported-assets": true
}
```

Not really intentional but it also removes the original STAC Collection and Item documents that `export_workspace` generates.

Attempting to download the original assets now returns a 500 error, as expected:

```
Traceback (most recent call last):
  File "/opt/openeo/lib/python3.8/site-packages/openeo_driver/views.py", line 1247, in _stream_from_s3
    s3_file_object = s3_instance.get_object(Bucket=bucket, Key=key, **dict_no_none(Range=bytes_range))
  File "/opt/openeo/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/opt/openeo/lib/python3.8/site-packages/botocore/client.py", line 676, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: None
```

bossie:
With regards to getting the `href`s right, I'm guessing just adding a `public_href` (an aggregator feature) to the exported assets and pointing this to the URI of the asset in the workspace will do the trick (and at the same time make the aggregator do the right thing as well): TBC

bossie:
Upon removal, the assets' `href`s are replaced with their workspace `s3:` or `file:` URIs.

As expected, attempting to download them with the Python client will fail, for example:

```
requests.exceptions.InvalidSchema: No connection adapters were found for 's3://<snip>/vdboschj-test3/openEO_2023-06-01Z.tif
```

Because of this and the fact that the documentation for `export_workspace` does not mention removal of the original assets, I'm leaving the job_option in.

PR #925, opened 2024-11-05
override asset href with workspace URI #883

bossie:
Points an asset's `public_href` to the one returned from the workspace so that it replaces the default download URL in the STAC Collection/Item.


bossie:
Things to consider:

- currently limited to [`DiskWorkspace`](https://github.com/Open-EO/openeo-python-driver/pull/321) so support `ObjectStorageWorkspace` as well;
- a `save_result` with multiple `export_workspace` and support for "alternate" hrefs;
- there's a bidirectional mapping in there to cope with the fact that a workspace is not concerned with assets, only files/objects; maybe the logic has to be extracted from `SaveResult#export_workspace`. The fact that this method accepts a `WorkspaceRepository` is probably a symptom of this.

PR #929, opened 2024-11-06
Test run job to s3

EmileSonneveld:
This was used to debug some batch_job to S3 issues. 
run_local_graph was made more robust
and formatted some files with black


openeogeotrellis/utils.py
```diff
@@ -442,6 +442,8 @@ def single_value(xs):
 def add_permissions(path: Path, mode: int):
     # TODO: accept PathLike etc as well
     # TODO: maybe umask is a better/cleaner option
+    if str(path).lower().startswith("s3:/"):
+        return
```
EmileSonneveld, line 446: @JeroenVerstraelen do you think it makes sense to skip add_permission on s3 URLs? 
The path.exists() will always give false

EmileSonneveld: @soxofaan This is where there are accidentally empty folders created when using S3 urls

soxofaan: How can that `if` make a difference? That `path` variable (being a `Path` object) will never start with "s3:/" because that is not a vailid relative/absolute posix path, right?

EmileSonneveld: Sometimes s3 paths do get loaded as `Path`. In those cases, the prefix loses a slash tough. (`s3://` becomes `s3:/`)
That's why I only use one slash in the `startswith` check.

Idealy this should be some kind of URI object instead of a path tough, because the double slash seems to be interpreted as redundant and is removed

```python
str(pathlib.Path("s3://test.png"))
-> 's3:/test.png'
```

PR #934, opened 2024-11-07
Filepath per band

EmileSonneveld:
The keys in stac data are relative paths to assets when they are specified in sub folders.
This is only visible where nfilepath_per_band is used ofc
```json
...
"assets": {
    "folder1/lon.tif": {
        "href": "/tmp/pytest-of-emile/pytest-77/test_filepath_per_band_False_0/folder1/lon.tif",
        "type": "image/tiff; application=geotiff",
        ...
    }
}
...
```


EmileSonneveld:
Will merge as it does not have a high impact. Any changes can always be done later

openeogeotrellis/deploy/batch_job.py
```diff
@@ -509,11 +509,10 @@ def write_metadata(metadata, metadata_file, job_dir: Path):
             s3_instance = s3_client()
 
             logger.info("Writing results to object storage")
-            for filename in os.listdir(job_dir):
-                if filename == UDF_PYTHON_DEPENDENCIES_FOLDER_NAME:
-                    logger.warning(f"Omitting {filename} as the executors will not be able to access it")
+            for file_path in filter(lambda x: x.is_file(), job_dir.rglob("*")):
```
soxofaan, line 512: As discussed: the change from simple listing to recursive glob might have unintended impact. Note that `job_dir` might contain more than just the batch job results a user is interested in, e.g. a large file tree of UDF dependencies or other intermediate/auxiliary resources

PR #945, opened 2024-11-20
Log lock

EmileSonneveld:


openeogeotrellis/integrations/gdal.py
```diff
@@ -105,6 +105,33 @@ def to_dict(self) -> dict:
...
+
+    if pool_size == 1:
+        # no need for error-prone multiprocessing here (Typical for NetCDF output)
+        results = [callback(*arg_tuple) for arg_tuple in argument_tuples]
+    else:
+        pool = multiprocessing.Pool(pool_size)
+        job = [pool.apply_async(callback, arg_tuple, error_callback=error_handler) for arg_tuple in argument_tuples]
+        pool.close()
+        try:
+            results = [j.get(timeout=60) for j in job]
```
bossie, line 122:
```suggestion
        jobs = [pool.apply_async(callback, arg_tuple, error_callback=error_handler) for arg_tuple in argument_tuples]
        pool.close()
        try:
            results = [job.get(timeout=60) for job in jobs]
```

openeogeotrellis/integrations/gdal.py
```diff
@@ -105,6 +105,33 @@ def to_dict(self) -> dict:
...
+            pool.join()
+        except multiprocessing.context.TimeoutError:
```
bossie, line 124:
```suggestion
        except multiprocessing.TimeoutError:
```

bossie (changes requested):
Is it possible to reliably provoke a deadlock in your test? I haven't witnessed one yet in my runs.

Otherwise fine.

EmileSonneveld:
When disabling the 2 safeties in gdal.py, that added test will get in a deadlock. This build demonstrates that: https://jenkins.vgt.vito.be/job/openEO/job/openeo-geopyspark-driver/job/log_lock/2/

openeogeotrellis/integrations/gdal.py
EmileSonneveld: Works fine

bossie:
I get that but AFAICT your test currently circumvents the problem because there's only 1 output asset and it will therefore not use multiprocessing.

If I "disable" the `if pool_size == 1:` check in the code and run your test, it will succeed and print your deadlock warning, so that's good.

Is it possible to:
- increase the number of output assets to provoke multiprocessing by using a bigger temporal interval and GTiff output (this will likely need TestCollection-LonLat16x16 instead);
- then check for the deadlock warning in stdout?

Issue #940, opened 2024-11-14
batch_job.py `write_metadata`: avoid ad-hoc file selection for upload


soxofaan:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/b63280c5fc5b928a0d231ab1aec6e3b47b4b9c36/openeogeotrellis/deploy/batch_job.py#L511-L519

Here we're building an ugly ad-hoc deny-list for "files"  that should not be uploaded to S3

As mentioned in the TODO, we should use an explicit asset list to upload instead of blindly assuming everything from the job dir should be uploaded (minus  some hand-picked exceptions)

soxofaan:
cc @EmileSonneveld

soxofaan:
As an illustration that this does not scale:
```
if UDF_PYTHON_DEPENDENCIES_FOLDER_NAME in str(file_path):
```

doesn't even work as `UDF_PYTHON_DEPENDENCIES_FOLDER_NAME` is not in play anymore on CDSE

EmileSonneveld:
This issue is a direct result of changes introduced in https://github.com/Open-EO/openeo-geopyspark-driver/issues/877

Issue #926, opened 2024-11-05
avoid using local paths as "href"


soxofaan:
in various places (mainly around batch job result/asset handling) we are putting local (absolute) paths in a "href" field. This is quite confusing as there is no indication that this is about local paths (e.g. at least a "file://" protocol would make that more explicit). It's also not ideal that  at some point that "href" probably has to be overwritten with  a relative/absolute (HTTP/S3) URL to be usable by end user and the original location is lost. Overall it makes it hard to reason about how assets are handled in our code

bossie:
It's clear from the STAC spec that "href" should be a URI with a scheme so I would even call this a bug.

Re: technical debt I would also avoid passing URIs around as strings and instead use a dedicated type as it will probably get rid of a great deal of implicit assumptions and workarounds like (think // vs /).

I haven't found a dedicated URI class in Python like `java.net.URI`; maybe we could look into https://pypi.org/project/uri/. Automatic translation between the two by py4j should be trivial.

soxofaan:
another symptom: https://github.com/Open-EO/openeo-geopyspark-driver/pull/929#discussion_r1838256145

PR #946, opened 2024-11-25
Fixed upload metadata

EmileSonneveld:
Only upload the relevant metadata files instead of the whole job-directory.
After merging, CDSE integration tests should run once with fusemount disabled.
https://github.com/Open-EO/openeo-geopyspark-driver/issues/940

### DIFF
--- a/openeogeotrellis/deploy/batch_job.py
+++ b/openeogeotrellis/deploy/batch_job.py
@@ -363,7 +363,7 @@ def run_job(
             ml_model_metadata=ml_model_metadata,
         )
         # perform a first metadata write _before_ actually computing the result. This provides a bit more info, even if the job fails.
-        write_metadata({**result_metadata, **_get_tracker_metadata("")}, metadata_file, job_dir)
+        write_metadata({**result_metadata, **_get_tracker_metadata("")}, metadata_file)
 
         for result in results:
             result.options["batch_mode"] = True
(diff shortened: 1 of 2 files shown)


openeogeotrellis/deploy/batch_job.py
```diff
@@ -270,6 +270,7 @@ def run_job(
 ):
     result_metadata = {}
 
+    stac_hrefs = []
```
soxofaan, line 273: If I understand correctly, this list is for collecting local paths. Using "href" in the name of this new variable is then not ideal as it  adds to the confusion between real hrefs and local paths as discussed in:
- https://github.com/Open-EO/openeo-geopyspark-driver/issues/926

EmileSonneveld: Is `stac_local_hrefs` a better name?
Href can be a s3 url, a https link, or a `file://paths` . 
At this point in the code it is still a filepath. 
After `_convert_job_metadatafile_outputs_to_s3_urls`, it becomes an s3 url.

soxofaan: if that list contains local paths, I would avoid "href" and just call it `stac_paths`

I'm not that familiar with all the code in play here, but if those values are changed later to real hrefs, that's fine, but just switch to a different variable name only then

Issue #648, opened 2024-01-23
load_stac: pixel shift


jdries:
load_stac has to use proj:bbox or proj:geometry to determine the exact bounding box in the original CRS.

This then allows us to construct results without pixel shift.

This line takes care of aligning global_extent to the collection metadata, and won't work for load_stac, hence the issue:
https://github.com/Open-EO/openeo-python-driver/blob/182ead3c19b8971e420929f71977d0ac9d1eaa26/openeo_driver/ProcessGraphDeserializer.py#L508

Proposed solution: store collection metadata from load_stac in the EvalEnv, making it available everywhere we might need it.

EmileSonneveld:
Another clear example: https://discuss.eodc.eu/t/spatial-shift-when-loading-data-from-the-stac-catalogue/738/3

jdries:
needs to be tested again because load_stac received a quite big rework

EmileSonneveld:
There is still a shift on at least CDSE staging and prod on Valentinas example:
![Image](https://github.com/user-attachments/assets/9fb200a4-f0c5-451b-8fdd-ce6d54e2a90f)

But the local batchjob test that loads OpenEO results as stac and saves it again has no shift. It might be due to a difference in stac metadata.

The `ERA5-Land-monthly-averaged-data` collection in ANIN that has metadata from the stac-builder has the offset.

jdries:
@EmileSonneveld  job id of most recent job with the shift?

EmileSonneveld:
era5 monthly: `j-241017ea45734856a92277e463781f40`
I compared the result with the source tiffs in `/data/users/...` in Q-GIS to test

EmileSonneveld:
In `SCF_20230120000000.tif` the origin is not a multiplication of the pixel size.
```
Origin	        631800.00,5184200.00
Pixel Size	500,-500
```
When resampling the pixel size to 100, the offset is gone. So probably there is a pixel perfect alignment happening in openEO that causes this shift.
`datacube = datacube.resample_spatial(resolution=100, projection="EPSG:32632")`

EmileSonneveld:
`load_collection` uses `cube:dimensions` to align the output result pixel grid with the one of the collection.

[...]

PR #953, opened 2024-11-28
Pixel shift

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/648

The stac catalog in this PR can be loaded with this URL: https://raw.githubusercontent.com/Open-EO/openeo-geopyspark-driver/refs/heads/pixel_shift/tests/data/stac/issue648-pixel-shift/collection.json 

This MR failed to build a few times, but started working again by itself. It failed and worked on jenkinsslave7.


openeogeotrellis/load_stac.py
```diff
@@ -496,6 +496,10 @@ def intersects_temporally(interval) -> bool:
     metadata_properties = {}
     correlation_id = env.get(EVAL_ENV_KEY.CORRELATION_ID, "")
 
+    if load_params.global_extent is None or len(load_params.global_extent) == 0:
+        # Layer catalog global extent is not known at the moment load_params is created.
+        load_params = load_params.copy()
+        load_params.global_extent = stac_bbox.as_dict()
```
EmileSonneveld, line 502: @bossie A better way would be to do a dry-run to create a new load_params object. Or a virtual layer in the layer catalog. 
Would this be an ok solution for now?

EmileSonneveld: global_extent is needed when calling `datacube_parameters.create()` in the next line. It is used for pixel perfect alignment with source data

EmileSonneveld:
The stac catalog in this PR can be loaded with this URL: https://raw.githubusercontent.com/Open-EO/openeo-geopyspark-driver/refs/heads/pixel_shift/tests/data/stac/issue648-pixel-shift/collection.json

openeogeotrellis/load_stac.py
jdries: but how can there be alignment if we just use the local stac bbox?
My suggestion was indeed to properly track the stac collection bounding boxes during dry run, allowing to do a real global extent construction, just like we do for load_collection.

EmileSonneveld: global_extent for load_stac is not read from cube:dimensions

PR #955, opened 2024-11-29
workspaces: allow adding a region name in the config

pvbouwel:
This does not perform a functional change yet but it does alow specifying a region name as part of the workspace config. That way configs can be updated after this version is life and subsequently a change can be deployed that depends on the region being defined.


bossie (changes requested):
OK implementation-wise. Please add an entry to the CHANGELOG.

Issue #906, opened 2024-10-16
gdal.py can not use logging library with default multiprocessing


JeroenVerstraelen:
**Description:**

We've encountered an issue with deadlocks arising from an interaction between Python's multiprocessing and logging libraries in gdal.py. These deadlocks occur sporadically, leading to batch jobs running significantly longer than expected or getting stuck entirely.

**Problem:**
- The root cause appears to be the sharing of memory between processes in the multiprocessing library, which happens by default. This sharing can lead to deadlocks in certain scenarios.
- As an initial fix, we disabled memory sharing, but this caused occasional Out of Memory errors.
- For now, we’ve worked around this issue by removing the logging functionality to prevent deadlocks.

**Next Steps:**
The ultimate solution will involve refactoring the module to reduce the scope of multithreading, but this will require more extensive changes.

**Current Workaround:**
Temporarily, we've opted to remove logging to avoid the deadlocks until we can rewrite the multithreaded parts of the module.

Commit that turned off logging in gdal.py:
https://github.com/Open-EO/openeo-geopyspark-driver/commit/c1b8676b3c93183082c93b97601cabe38e643f34

bossie:
Restored a couple of log entries (poorly) in the context of https://github.com/eu-cdse/openeo-cdse-infra/issues/278; they use `print()` rather than a logger.

EmileSonneveld:
A user encountered this issue here: https://forum.dataspace.copernicus.eu/t/sentinel-3-syn-gets-stuck/1429/3

To fix it, it is probably best to create the thread pool with `get_context("spawn")` instead of the default way that uses forks. (The fork way appears to copy mutexes, that could be in their locked state. When the parent process unlocks them, the child process would still have a locked mutex)

On top of that, it would be nice to add a timeout to the child processes.

bossie:
I figured that would happen as logs are not limited to our own code.

EmileSonneveld:
Gdal info got stuck even without multi threading:
`kubectl ... logs a-XXX-driver`
```python
...
[...]

PR #960, opened 2024-12-03
Pass environment variables to Spark

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/906


EmileSonneveld:
better in GpsBackendConfig

Issue #964, opened 2024-12-04
Logging from run_udf on vectorcube is not visible


EmileSonneveld:
Found while checking this issue: 
https://discuss.eodc.eu/t/preflight-process-graph-validation-raised/1026

EmileSonneveld:
Works on CDSE prod

PR #965, opened 2024-12-04
Vectorcube udf logging

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/964


openeogeotrellis/udf.py
```diff
@@ -26,6 +27,7 @@
 UDF_PYTHON_DEPENDENCIES_ARCHIVE_NAME = "udf-py-deps.zip"
 
 
+@ensure_executor_logging
```
EmileSonneveld, line 30: @bossie Could ensure_executor_logging be a performance problem here? It also gets called from the driver, so does not need the attribute in every case. But it is handy to just put it on one location

tests/test_udf.py
```diff
@@ -380,3 +384,49 @@ def apply_udf_data(data: UdfData):
...
+
+        run_job(
+            {"process_graph": post_processed.flat_graph()},
+            output_file=tmp_path / "out",
+            metadata_file=tmp_path / JOB_METADATA_FILENAME,
+            api_version="2.0.0",
+            job_dir=tmp_path,
+            dependencies=[],
+        )
+        # TODO: Use caplog instead
```
EmileSonneveld, line 431: @soxofaan Caplog does not seem to take the executor logs. 
The workaround seems more complicated than just reading the logfile:
https://github.com/pytest-dev/pytest/issues/3037#issuecomment-595466044 
Ok to keep like this?

openeogeotrellis/udf.py
EmileSonneveld: (Build currently failing due to error on master)

tests/test_udf.py
```diff
@@ -380,3 +384,49 @@ def apply_udf_data(data: UdfData):
...
+import logging
+def apply_udf_data(data: UdfData):
+    logging.warn('Hello from UDF!')
+    return data
+"""
+        post_processed = vector_cube.process("run_udf", data=vector_cube, udf=udf, runtime="Python")
+
+        log_dir = os.environ.get("LOG_DIRS", ".").split(",")[0]
+        log_file = Path(log_dir) / "openeo_python.log"
+        open(log_file, "w").close()  # Clear file content
```
soxofaan, line 421: how does this work?  there seem to be a lot of brittle assumptions here about `LOG_DIRS` and `openeo_python.log`.  Can't we just make sure that the logging is done to a file under `tmp_path`?

This also looks pretty evil:
```
open(log_file, "w").close() # Clear file content
```

you are blindly clearing a file, possibly/probably in current working directory. That's not something that tests should do

openeogeotrellis/udf.py
soxofaan: note that `ensure_executor_logging` only does stuff the first time the decorated function is called, after that it's basically a no-op I think

tests/test_udf.py
```diff
@@ -380,3 +384,49 @@ def apply_udf_data(data: UdfData):
...
+                    [5.07, 51.215],
+                    [5.07, 51.22],
+                    [5.08, 51.22],
+                    [5.08, 51.215],
+                    [5.07, 51.215],
+                ]
+            ],
+        }
+
+        vector_cube = datacube.aggregate_spatial(geometries=geometries, reducer="mean")
```
soxofaan, line 408: If you just want to test logging from UDF's I don't think you need all this load_stac/local_batch_job/aggregate_spatial setup. Eventually you just want a simple process graph, with "top level" run_udf process.

Ultimately I even think it should be possible to just work with this minimal, no-cubes, process graph:

```json
{
  "runudf1": {
    "process_id": "run_udf",
    "arguments": {
      "data": [1, 2, 3],
      "runtime": "Python",
      "udf": "UDF CODE HERE"
    }
  }
}
```

tests/test_udf.py
soxofaan: sure, if the logging is done from a subprocess, `caplog` won't capture that, so fine to read it straight from file.

However to be correct in this case I think, the UDF is executed in driver process, not executors. But regardless it's still a (driver) subprocess I think

tests/test_udf.py
EmileSonneveld: That works too, awesome!

tests/test_udf.py
EmileSonneveld: I made the message unique, so that the logfile does not need to be cleared first.

I had no direct success with `monkeypatch.setenv("LOG_DIRS", str(tmp_path))`
It does not affect the logging path https://github.com/Open-EO/openeo-python-driver/blob/0e00182d000d052603053632ea5d3c347baf3ba9/openeo_driver/util/logging.py#L88

EmileSonneveld: Changed hardcoded path to:
```python
get_logging_config()["handlers"][LOG_HANDLER_FILE_JSON]["filename"]
```
Seems better to me. Ok to merge like this?

tests/test_udf.py
EmileSonneveld: The code is called here in the executors: 
https://github.com/Open-EO/openeo-geopyspark-driver/blob/289b6b1e64a31300958c525ba037b2ca39938d2b/openeogeotrellis/backend.py#L1259

PR #969, opened 2024-12-10
feature: provide s3 profiles and tokens during execution

pvbouwel:
When a job gets submitted generate an identity token which can be exchanged for AWS credentials that can be used against the S3 proxy but scoped to the job context.

Additionally a config file gets generated with AWS Profiles which can then be used in order to access S3 via the S3 proxy. This way using an AWS profile is the only config needed for your S3 access. The profile names correspond to workspace names for workspace access.

For normal usage we add a:
 - base profile which overtime can be used for access to OpenEO platform resources (e.g. results bucket, logs, ...)
 - eodata profile which allows access to the EOdata collection.

This change is behind a config feature flag which defaults to false to allow gradual rollout and testing.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -81,7 +81,7 @@ from openeogeotrellis.config.s3_config import S3Config
 from openeogeotrellis.configparams import ConfigParams
 from openeogeotrellis.geopysparkdatacube import GeopysparkCubeMetadata, GeopysparkDataCube
 from openeogeotrellis.integrations.etl_api import get_etl_api, ETL_ORGANIZATION_ID_JOB_OPTION
-from openeogeotrellis.identity import IDP_TOKEN_ISSUER
+from openeogeotrellis.integrations.identity import IDP_TOKEN_ISSUER
 from openeogeotrellis.integrations.hadoop import setup_kerberos_auth
 from openeogeotrellis.integrations.kubernetes import (
     k8s_job_name, kube_client,
@@ -2045,7 +2045,7 @@ class GpsBatchJobs(backend.BatchJobs):
                     "batch_job_cfg_secret.yaml.j2",
                     secret_name=batch_job_cfg_secret_name,
                     job_id=job_id,
-                    token=IDP_TOKEN_ISSUER.get_identity_token(user_id, job_id),
+                    token=IDP_TOKEN_ISSUER.get_job_token(user_id, job_id),
                     profile_file_content=S3Config.from_backend_config(job_id, "/opt/job_config/token")
                 )
 
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -2037,17 +2037,19 @@ class GpsBatchJobs(backend.BatchJobs):
                     "OPENEO_STAC_OIDC_CLIENT_SECRET_STAC_OPENEO_DEV"
                 ),
                 provide_s3_profiles_and_tokens=get_backend_config().provide_s3_profiles_and_tokens,
-                batch_job_cfg_secret_name=batch_job_cfg_secret_name
+                batch_job_cfg_secret_name=batch_job_cfg_secret_name,
+                batch_job_config_dir=get_backend_config().batch_job_config_dir
             )
 
             if get_backend_config().provide_s3_profiles_and_tokens:
                 # For now we cannot access the subject of the initial access token but generally it is the user_id
+                token_path = get_backend_config().batch_job_config_dir / "token"
                 s3_profiles_cfg_batch_secret = k8s_render_manifest_template(
                     "batch_job_cfg_secret.yaml.j2",
                     secret_name=batch_job_cfg_secret_name,
                     job_id=job_id,
                     token=IDP_TOKEN_ISSUER.get_job_token(sub_id=user_id, user_id=user_id, job_id=job_id),
-                    profile_file_content=S3Config.from_backend_config(job_id, "/opt/job_config/token")
+                    profile_file_content=S3Config.from_backend_config(job_id, str(token_path))
                 )
 
             if get_backend_config().fuse_mount_batchjob_s3_bucket:
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -77,9 +77,11 @@ import numpy as np
 import openeogeotrellis
 from openeogeotrellis import sentinel_hub, load_stac, datacube_parameters
 from openeogeotrellis.config import get_backend_config
+from openeogeotrellis.config.s3_config import S3Config
 from openeogeotrellis.configparams import ConfigParams
 from openeogeotrellis.geopysparkdatacube import GeopysparkCubeMetadata, GeopysparkDataCube
 from openeogeotrellis.integrations.etl_api import get_etl_api, ETL_ORGANIZATION_ID_JOB_OPTION
+from openeogeotrellis.identity import IDP_TOKEN_ISSUER
 from openeogeotrellis.integrations.hadoop import setup_kerberos_auth
 from openeogeotrellis.integrations.kubernetes import k8s_job_name, kube_client, truncate_job_id_k8s, k8s_render_manifest_template
 from openeogeotrellis.integrations.traefik import Traefik
@@ -2027,7 +2029,17 @@ class GpsBatchJobs(backend.BatchJobs):
                 openeo_stac_oidc_client_secret_stac_openeo_dev=os.environ.get(  # TODO: pass a list or dict?
                     "OPENEO_STAC_OIDC_CLIENT_SECRET_STAC_OPENEO_DEV"
                 ),
+                provide_s3_profiles_and_tokens=get_backend_config().provide_s3_profiles_and_tokens
             )
+
+            if get_backend_config().provide_s3_profiles_and_tokens:
+                s3_profiles_cfg_batch_secret = k8s_render_manifest_template(
+                    "s3_access_cfg_batch_job.yaml.j2",
+                    job_name=spark_app_id,
+                    job_id=job_id,
+                    token=IDP_TOKEN_ISSUER.get_identity_token(user_id, job_id),
+                    profile_file_content=S3Config.from_backend_config(job_id, "/opt/job_config/token")
+                )
 
             if get_backend_config().fuse_mount_batchjob_s3_bucket:
                 persistentvolume_batch_job_results_dict = k8s_render_manifest_template(
@@ -2051,6 +2063,8 @@ class GpsBatchJobs(backend.BatchJobs):
                     if get_backend_config().fuse_mount_batchjob_s3_bucket:
                         api_instance_core.create_persistent_volume(persistentvolume_batch_job_results_dict, pretty=True)
                         api_instance_core.create_namespaced_persistent_volume_claim(pod_namespace, persistentvolumeclaim_batch_job_results_dict, pretty=True)
+                    if get_backend_config().provide_s3_profiles_and_tokens:
+                        api_instance_core.create_namespaced_secret(pod_namespace, s3_profiles_cfg_batch_secret, pretty=True)
                     submit_response_sparkapplication = api_instance_custom_object.create_namespaced_custom_object("sparkoperator.k8s.io", "v1beta2", pod_namespace, "sparkapplications", sparkapplication_dict, pretty=True)
                     log.info(f"mapped job_id {job_id} to application ID {spark_app_id}")
                     dbl_registry.set_application_id(job_id, user_id, spark_app_id)
@@ -2823,6 +2837,28 @@ class GpsBatchJobs(backend.BatchJobs):
                             # TODO: more precise checking that it was indeed the app that was not found (instead of k8s api itself).
                             logger.info(
                                 f"The storage resources for {application_id} could not be found."
+                            )
+
+                if get_backend_config().provide_s3_profiles_and_tokens:
+                    secret_name = f's3-{application_id}'
+                    try:
+                        delete_response_secret = api_instance_core.delete_namespaced_secret(
+                            name=secret_name,
+                            namespace=namespace,
+                            pretty=True
+                        )
+                        logger.debug(
+                            f"Removed secret {secret_name} with kubernetes API call",
+                            extra={'job_id': job_id, 'API response': delete_response_secret}
+                        )
+                    except kubernetes.client.exceptions.ApiException as e:
+                        if e.status == 404:
+                            logger.info(
+                                f"The secret {secret_name} could not be found."
+                            )
+                        else:
+                            logger.warning(
+                                f"Got exception {e} when deleting secret {secret_name}"
                             )
 
                 with self._double_job_registry:
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -83,7 +83,12 @@ from openeogeotrellis.geopysparkdatacube import GeopysparkCubeMetadata, Geopyspa
 from openeogeotrellis.integrations.etl_api import get_etl_api, ETL_ORGANIZATION_ID_JOB_OPTION
 from openeogeotrellis.identity import IDP_TOKEN_ISSUER
 from openeogeotrellis.integrations.hadoop import setup_kerberos_auth
-from openeogeotrellis.integrations.kubernetes import k8s_job_name, kube_client, truncate_job_id_k8s, k8s_render_manifest_template
+from openeogeotrellis.integrations.kubernetes import (
+    k8s_job_name, kube_client,
+    truncate_job_id_k8s,
+    k8s_render_manifest_template,
+    k8s_get_batch_job_cfg_secret_name,
+)
 from openeogeotrellis.integrations.traefik import Traefik
 from openeogeotrellis.job_registry import (
     DoubleJobRegistry,
@@ -1967,6 +1972,8 @@ class GpsBatchJobs(backend.BatchJobs):
 
             image_name = job_options.get("image-name", running_image)
 
+            batch_job_cfg_secret_name = k8s_get_batch_job_cfg_secret_name(spark_app_id)
+
             sparkapplication_dict = k8s_render_manifest_template(
                 "sparkapplication.yaml.j2",
                 job_name=spark_app_id,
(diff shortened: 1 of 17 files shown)


openeogeotrellis/identity.py
```diff
@@ -0,0 +1,124 @@
...
+    def get_identity_token(self, user_id: str, job_id: str) -> Optional[str]:
+        self._reload_idp_details_if_needed()
+        if self._IDP_DETAILS is None:
+            return None
+        else:
+            idp_details = self._IDP_DETAILS
+            now = utcnow()
+            return jwt.encode(
+                {
+                    "sub": user_id,  # TODO: see if we can use subject from upstream IDP instead
```
pvbouwel, line 94: @soxofaan I wanted to check this with you before having to dive to deep in the code. In a way it is OK to have user_id because that is what we use to identify users in web driver and during execution. But since the user_id is also a claim in the token it could also make sense to keep the subject of the initial token (so the openeo access token) but I am not sure whether that is easily retrievable at this stage.

setup.py
```diff
@@ -98,7 +98,8 @@
         "planetary-computer~=1.0.0",
         "reretry~=0.11.8",
         "traceback-with-variables==2.0.4",
-        'scipy>=1.8' # used by sentinel-3 reader
+        'scipy>=1.8',  # used by sentinel-3 reader
+        "PyJWT[crypto]~=2.9.0",  # For identity tokens
```
pvbouwel, line 102: >= 2.10 dropped support for Python 3.8

soxofaan: But we want to upgrade to py 3.11 eventually, were there is no reason I guess to stick to 2.9.0?
can't we leave that contraint open, or just specify `>=2.9.0`?

openeogeotrellis/identity.py
soxofaan: I guess you actually need the "sub" identifier here, which is the same as user_id in most cases I guess, but I'm not sure we can guarantee that (I also think we have a ticket somewhere to actually decouple the user_id from the "sub" identifier). So I think it's best to be explicit here and work with a variable "sub_id" here instead of "user_id".   

Another reason to avoid "user_id" here are client credential flows where the sub id is associated with the client, and not the user

openeogeotrellis/integrations/kubernetes.py
```diff
@@ -76,16 +79,25 @@ def k8s_state_to_openeo_job_status(state: str) -> str:
...
     )
     jinja_dir = os.path.dirname(jinja_path)
-    jinja_template = Environment(
+    jinja_env = Environment(
         loader=FileSystemLoader(jinja_dir)
-    ).from_string(open(jinja_path).read())
+    )
+
+    def base64encode(input_str: str) -> str:
+        return base64.b64encode(input_str.encode()).decode()
```
soxofaan, line 96:
```suggestion
        return base64.b64encode(input_str.encode("utf8")).decode("utf8")
```

just to be sure

tests/test_identity.py
```diff
@@ -0,0 +1,115 @@
...
+              "public_key": pem_public_key,
+            },
+            fh
+        )
+    return pem_public_key
+
+
+@pytest.fixture
+def idp_token_issuer(monkeypatch) -> Generator[str, None, None]:
+    with tempfile.TemporaryDirectory(
```
soxofaan, line 60: use the `tmp_path` fixture instead to automatically take care of all temp folder management

e.g. you just would need something like 

```python
@pytest.fixture
def idp_token_issuer(monkeypatch, tmp_path) -> Generator[str, None, None]:
    idp_cfg_file = tmp_path / "idp.json"
    monkeypatch.setattr(IDP_TOKEN_ISSUER, '_IDP_DETAILS_FILE', idp_cfg_file)
    public_key = create_idp_config(idp_cfg_file)
    yield public_key
```

tests/test_identity.py
```diff
@@ -0,0 +1,115 @@
...
+def test_idptokenissuer_should_not_be_instantiated_manually():
+    # WHEN someone wants to manually create an instance of the IDPTokenIssuer
+    # THEN they should get an exception
+    with pytest.raises(RuntimeError):
+        IDPTokenIssuer()
+
+
+def test_without_valid_config_token_should_just_be_none(monkeypatch):
+    # GIVEN no valid config
+    monkeypatch.setattr(IDP_TOKEN_ISSUER, '_IDP_DETAILS_FILE', Path("/tmp/file_that_do3s_not_exist.arr"))
```
soxofaan, line 24: Isn't it possible to set `_IDP_DETAILS_FILE` from config (instead of from env var)? then it would be possible to override it in tests without monkeypatching

openeogeotrellis/identity.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import annotations
```
soxofaan, line 1: "identity" seems like a very generic name, while this is specifically about AWS user and token management?
If so, I think it's cleaner to move this somewhere under the "integrations" namespace, instead of making this a top-level module.

In the longer term I'd like to slim down the whole top-level of openeogeotrellis and make things more modular/isolated

openeogeotrellis/backend.py
```diff
@@ -2819,6 +2833,28 @@ def cancel_job(self, job_id: str, user_id: str):
                                 f"The storage resources for {application_id} could not be found."
                             )
 
+                if get_backend_config().provide_s3_profiles_and_tokens:
+                    secret_name = f's3-{application_id}'
```
soxofaan, line 2837: Isn't it better to get the secret_name from where it was created, instead of regenerating it here.

if I understand correctly it comes from s3_access_cfg_batch_job.yaml.j2  but that is not obvious and consequently a brittle, hidden coupling

soxofaan (commented):
that was a lot to read through :)

a couple of review notes

openeogeotrellis/identity.py
pvbouwel: I can modify this function to take sub_id but I don't know how/where to get the sub identifier in this openeo-geopyspark-driver module (particularly during the start job method in backend.py).

setup.py
pvbouwel: I thought to pin it to avoid breaking it with updates but that is not a good reason to pin. My unittests should catch issues for breaking due to upgrades so I'll go for the `>=` suggestion

openeogeotrellis/integrations/kubernetes.py
pvbouwel: I was under the impression that utf-8 was the default but I don´t mind explicit so will change

setup.py
soxofaan: indeed, I'm not a fan of putting too much constraints or hard pins in setup.py/pyproject.toml. In the long term you risk problems with unsolvable dependencies, impossible upgrades, or worse, security issues.

If there is clear need for pins that should go into a deploy `requirements.txt`

openeogeotrellis/integrations/kubernetes.py
soxofaan: The default encoding depends on env vars and that kind of black magic. :magic_wand: 
So to avoid endless "works on my computer" debug sessions, I try to always be explicit about it

openeogeotrellis/identity.py
soxofaan: yes, I would make that new function future-proof and generic with a honest argument name "sub_id".

And where you call it you can then make it more explicit that we are doing a bit of user_id->sub_id mapping, which is fine for now:
```
.get_identity_token(sub_id=user_id, job_id=job_id)
```

openeogeotrellis/identity.py
pvbouwel: Identity is indeed quite generic but that was on purpose. At this stage we will indeed only use it for AWS like interactions (S3 access via S3 proxy) but there is no reason why this would not be applicable to any other type of services. If we have additional services that we build and deploy this identity can also be used to call those services where this identity will have assertions/claims. I don´t mind moving it to integrations module though. Perhaps naming it job_identity or job_identity_token given that it is really tied to a job execution?

openeogeotrellis/backend.py
pvbouwel: I agree, I did an attempt on improving the situation. By having a function to get a secret name. I also stopped liking the secret name as it would make it confusing if in the future there is other config that needs to be passed on to the batch-jobs so made the naming more generic.
 
One thing I struggled with was were to put the function. In the end I put in under integrations/kubernetes because it only applies to kubernetes.

tests/test_identity.py
pvbouwel: I have moved it to config.

openeogeotrellis/integrations/kubernetes.py
pvbouwel: Aha TIL, good to know I can imagine indeed it being a "time sink" debug session to troubleshoot something like that

tests/integrations/test_identity.py
```diff
@@ -0,0 +1,113 @@
+import json
+import time
+from pathlib import Path
+from typing import Generator
+
+import jwt
+import pytest
+
+from cryptography.hazmat.primitives.asymmetric import rsa
```
soxofaan, line 9: I don't think `cryptography` is already an explicit dep in setup.py. I guess it works because it's a transitive dependency of PyJWT or something. But if we're importing it like that, I think we should make that an explicit dependency

soxofaan: Ok I didn't consider this was just a test dependency. This makes it indeed much less an important breaking point and I'm fine with leaving it implicit, which has advantages too as you noted

tests/integrations/test_identity.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def test_idptokenissuer_should_not_be_instantiated_manually():
+    # WHEN someone wants to manually create an instance of the IDPTokenIssuer
+    # THEN they should get an exception
+    with pytest.raises(RuntimeError):
+        IDPTokenIssuer()
+
+
+def test_without_valid_config_token_should_just_be_none(monkeypatch):
```
soxofaan, line 22: `monkeypatch` is not used, so do you have to declare it?

tests/integrations/test_identity.py
```diff
@@ -0,0 +1,113 @@
...
+              "private_key": pem_private_key,
+              "public_key": pem_public_key,
+            },
+            fh
+        )
+    return pem_public_key
+
+
+@pytest.fixture
+def idp_token_issuer(monkeypatch, tmp_path) -> Generator[str, None, None]:
```
soxofaan, line 59: `monkeypatch` is not used, so do you have to declare it

tests/test_s3_config.py
```diff
@@ -0,0 +1,72 @@
...
+testdata = [
+    # inputString, expectedChange
+    ("1", True),  # Too short
+    ("1" * 100, True),  # Too long
+    ("valid-session123", False),
+    ("invalid-ch$racter", True),
+]
+
+
+@pytest.mark.parametrize("in_str, requires_sanitization", testdata)
```
soxofaan, line 65: I usually find it cleaner to define the first argument explicitly as a list if it is  a list:
```suggestion
@pytest.mark.parametrize(["in_str", "requires_sanitization"], testdata)
```

tests/test_utils.py
```diff
@@ -413,3 +415,43 @@ def test_map_optional():
 
     assert map_optional(to_upper, None) is None
     assert map_optional(to_upper, "hello") == "HELLO"
+
+
+def test_get_file_reload_register_func_if_changed():
+    with tempfile.TemporaryDirectory(
```
soxofaan, line 421: can't you use `tmp_path` fixture here too?

soxofaan (commented):
went another time through this PR (superficially)

openeogeotrellis/utils.py
```diff
@@ -802,3 +808,66 @@ def wait_till_path_available(path: Path):
...
+                file_mode=0,
+                owner_uid=0,
+                owner_gid=0,
+            )
+
+
+_last_config_reload: dict[Path, FileChangeFingerprint] = {}
+
+
+def get_file_reload_register_func_if_changed(file_path: Path, resolve=True) -> Optional[Callable[[], None]]:
```
soxofaan, line 856: I also feel there is something off here with this helper, but I can't really put my finger on it.
Maybe it's because it depends on a global `_last_config_reload` so it only works properly when used from some kind of singleton (or single owner) context. Which means it might not belong yet in a generic utils module

openeogeotrellis/backend.py
```diff
@@ -2022,8 +2031,20 @@ def as_arg_element(dependency: dict) -> dict:
...
             )
 
+            if get_backend_config().provide_s3_profiles_and_tokens:
+                # For now we cannot access the subject of the initial access token but generally it is the user_id
+                s3_profiles_cfg_batch_secret = k8s_render_manifest_template(
+                    "batch_job_cfg_secret.yaml.j2",
+                    secret_name=batch_job_cfg_secret_name,
+                    job_id=job_id,
+                    token=IDP_TOKEN_ISSUER.get_job_token(sub_id=user_id, user_id=user_id, job_id=job_id),
+                    profile_file_content=S3Config.from_backend_config(job_id, "/opt/job_config/token")
```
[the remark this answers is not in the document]
pvbouwel: This file is provided by the same secret. Kubernetes will mount his secret and the token file will appear. I did however push another commit d919872b286af2e56651c85453bf3b88217ce5b2  because the `/opt/job_config` is actually used in multiple locations so better extracted in a config var.

openeogeotrellis/backend.py
```diff
@@ -2819,6 +2842,28 @@ def cancel_job(self, job_id: str, user_id: str):
...
+                    batch_job_cfg_secret_name = k8s_get_batch_job_cfg_secret_name(application_id)
+                    try:
+                        delete_response_secret = api_instance_core.delete_namespaced_secret(
+                            name=batch_job_cfg_secret_name,
+                            namespace=namespace,
+                            pretty=True
+                        )
+                        logger.debug(
+                            f"Removed secret {batch_job_cfg_secret_name} with kubernetes API call",
+                            extra={'job_id': job_id, 'API response': delete_response_secret}
```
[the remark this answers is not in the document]
pvbouwel: This was more Monkey see monkey do as this is how it is used for removing PV a bit higher up.

The message has the crucial information though so I would say it is ok to have to expand the document in Kibana to see these extra fields.

openeogeotrellis/config/config.py
```diff
@@ -245,3 +246,13 @@ class GpsBackendConfig(OpenEoBackendConfig):
...
+
+    """
+    Inject S3 profiles and tokens that allow S3 access scoped to the Job execution.      
+    """
+    provide_s3_profiles_and_tokens: bool = False
+
+    """
+    Config file to generate identity information from this Openeo Instance
+    """
+    openeo_idp_details_file: Path = Path("/opt/.idp/idp_details.json")
```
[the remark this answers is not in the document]
pvbouwel: On our infrastructure it will exist already. It is meant to be a default that would work for any environment but if for some reason in certain environments it causes issues it could be overridden. The responsibility of creating this file is however not part of the batch job driver but our infra deploy pipelines so it indeed just exists already.

openeogeotrellis/integrations/identity.py
```diff
@@ -0,0 +1,130 @@
...
+    _SINGLETON: Optional[IDPTokenIssuer] = None
+
+    def __init__(self):
+        if self.__class__._SINGLETON is not None:
+            raise RuntimeError("IDPTokenIssuer is meant as a singleton")
+        self._IDP_DETAILS: Optional[IDPDetails] = None
+        self.__class__._SINGLETON = self
+
+    @classmethod
+    def instance(cls) -> "IDPTokenIssuer":
```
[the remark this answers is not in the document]
pvbouwel: In most cases I actually did it to allow usage of normal types like dict rather than the Dict counterpart from typing but indeed once the import is there it is OK to get rid of the quotes, will do.

openeogeotrellis/integrations/identity.py
```diff
@@ -0,0 +1,130 @@
...
+IDP_TOKEN_ISSUER = IDPTokenIssuer.instance()
+
+
+if __name__ == '__main__':
+    aws_config_dir = os.environ.get("AWS_CONFIG_DIR", "/opt/spark/work-dir")
+    aws_config_path = Path(aws_config_dir)
+    aws_token_filename = os.environ.get("AWS_TOKEN_FILENAME", "token")
+    aws_config_path.mkdir(parents=True, exist_ok=True)
+    with open(aws_config_path.joinpath(aws_token_filename), 'w') as token_file:
+        token_file.write(IDP_TOKEN_ISSUER.get_job_token("0", "none"))
```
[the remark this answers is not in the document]
pvbouwel: If OK I want to leave it in there for allowing troubleshooting of drivers when needed. Basically running this python file on the driver should store a token in a local file which would not influence operations but would allow to do some basic tests with the generated token. (e.g. making sure infra dependencies are correct).

openeogeotrellis/integrations/identity.py
```diff
@@ -0,0 +1,130 @@
...
+    class method `instance() to get the singleton object`.
+    """
+    _TOKEN_EXPIRES_IN_SECONDS = 48 * 3600
+    _SINGLETON: Optional[IDPTokenIssuer] = None
+
+    def __init__(self):
+        if self.__class__._SINGLETON is not None:
+            raise RuntimeError("IDPTokenIssuer is meant as a singleton")
+        self._IDP_DETAILS: Optional[IDPDetails] = None
+        self.__class__._SINGLETON = self
```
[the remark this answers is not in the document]
pvbouwel: I prefer it to be a singleton for the time being. The singleton check is not a one-way door as it can be easily reverted. Having 2 issuers on the other end creates a bunch of problems (especially with well-known paths once it is preferred to public key material.

tests/integrations/test_identity.py
pvbouwel: Nope, oversight of refactoring will fix

tests/integrations/test_identity.py
pvbouwel: Nope, oversight of refactoring will fix

tests/test_utils.py
pvbouwel: yes, will change.

openeogeotrellis/utils.py
pvbouwel: Is your concern the shared state and how it could give issues if multiple observe the same file. Because that was indeed not tackled in the implementation. I refactored it to be a watcher that can be instantiated. This way it can be safely used by multiple components.

https://github.com/Open-EO/openeo-geopyspark-driver/pull/969/commits/c0d631d691fdd2e139a1887b8a3fc971c1566f93

I can equally well move it to the identity module but it really is utility code and not specific to identity code.

openeogeotrellis/integrations/identity.py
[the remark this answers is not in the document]
soxofaan: I'm still confused. The main defining feature I see here is to reload some data from a file if you observe it has been touched since last read. But I don't understand how that requires a singleton.

And related: why would we want to hardcode here that there can only be one issuer in the application? For example: in openEO auth we support multiple OIDC providers at the same time

openeogeotrellis/integrations/identity.py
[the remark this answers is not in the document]
pvbouwel: I wasn't aware of this behavior but it does make sense.

openeogeotrellis/integrations/identity.py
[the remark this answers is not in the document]
pvbouwel: The reloading of the config does not require it to be a singleton as long as there would be a watcher per instance. The reason why it is a singleton is because there should not be a need for multiple IDPTokenIssuer's and if there is it wouldn't be hard to allow it.

Yes we support different OIDC providers for auth such that users don´t have to register themselves with us but we should not create different identity providers for tokens that we issue ourselves. If there are tokens required for different purposes that would still be possible within this singleton by adding limitations in a standard way (e.g. allowing multiple types of tokens and adding an audience for certain type of tokens that should be limited to specific audiences) but the issuer should always be a reference to the OpenEO instance.

So given that that is logically 1 instance for a given runtime a singleton is preferred.

pvbouwel:
Rebased against master to not be against a version of master that does not pass the tests.

PR #992, opened 2025-01-10
Feature/more reliable zookeeper

pvbouwel:
This change will just create 1 KazooClient and keeps it running in the background. The KazooClient itself has logic to recover connectivity.

In order to deal with intermittent failure we also retry operations if they fail due to connectivity related issues. At this time the once that came from test scenario's were `kazoo.exceptions.SessionExpiredError` and `kazoo.exceptions.ConnectionLoss` but if other retryable exceptions are identified they could be easily added.


jdries (approved):

soxofaan (commented):
(apparently I still had these comments from some time ago)

openeogeotrellis/integrations/zookeeper.py
```diff
@@ -0,0 +1,155 @@
...
+            if not self._zk.connected:
+                if self._logger is not None:
+                    self._logger.debug(f"No active connection for zookeeper client ({self._zkhosts})")
+                raise ZookeeperClient.NoActiveConnectionException()
+
+    _wait_up_to_1minute_for_zookeeper_connection = retry(
+        exceptions=(NoActiveConnectionException,),
+        tries=600,
+        logger=None,
+    )(_wait_100ms_for_zookeeper_connection)
```
soxofaan, line 130: while reviewing this I got a bit lost in this `_wait_100ms_for_zookeeper_connection`/`_wait_up_to_1minute_for_zookeeper_connection` construction.

Can't it be simplified to a single helper like 

```python
def _wait_on_connected(self, timeout: float=60, sleep=0.1):
    end = time.time() + timeout
    while time.time() < end:
        if self._zk.connected:
            return
        if self._logger is not None:
            self._logger.debug(f"No active connection for zookeeper client ({self._zkhosts})")
        time.sleep(sleep)
    raise ZookeeperClient.NoActiveConnectionException()
```
    



(PS: I would avoid hardcoding implementation details like `100ms` and `1minute` in the function names to make it easier to fine-tune these when necessary)

soxofaan:
while reviewing it also started wondering if it wouldn't be easier to just add an option for reuse to 
https://github.com/Open-EO/openeo-geopyspark-driver/blob/4f57b9b53cc91759bd088419207460c06f63024d/openeogeotrellis/user_defined_process_repository.py#L83-L94

e.g. PoC:
- #1037

that approach would also allow to roll client reuse out gradually (e.g. with config or feature flags), instead of a a full on replacement as proposed in this PR #992

PR #1012, opened 2025-01-21
Chmod only added files. Works fine with gdalinfo objects and missing metadata

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/352

### DIFF
--- a/openeogeotrellis/deploy/run_graph_locally.py
+++ b/openeogeotrellis/deploy/run_graph_locally.py
@@ -10,6 +10,7 @@ from openeo.util import ensure_dir
 
 
 def run_graph_locally(process_graph, output_dir):
+    files_orig = set(output_dir.rglob("*"))
     output_dir = ensure_dir(output_dir)
     setup_environment(output_dir)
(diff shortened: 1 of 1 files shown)


openeogeotrellis/deploy/run_graph_locally.py
```diff
@@ -19,6 +19,7 @@ def run_graph_locally(process_graph, output_dir):
     process_graph = as_flat_graph(process_graph)
     if "process_graph" not in process_graph:
         process_graph = {"process_graph": process_graph}
+    files_orig = set(output_dir.glob("*"))
```
bossie, line 22: Does this glob work with files not in this output directory e.g. `filepath_band_band`?

EmileSonneveld: Good question, I'll test

EmileSonneveld: A fix was needed indeed, Thanks!

PR #1037, opened 2025-02-04
PoC to introduce KazooClient reuse in user_defined_process_repository

soxofaan:
Proof of concept for eu-cdse/openeo-cdse-infra#325  to add an option for KazooClient reuse at
https://github.com/Open-EO/openeo-geopyspark-driver/blob/4f57b9b53cc91759bd088419207460c06f63024d/openeogeotrellis/user_defined_process_repository.py#L83-L94

this would be a simpler and configurable alternative for #992

### DIFF
--- a/openeogeotrellis/user_defined_process_repository.py
+++ b/openeogeotrellis/user_defined_process_repository.py
@@ -1,16 +1,15 @@
 import contextlib
 import json
 import logging
-from typing import List, Dict
-from typing import Union
+from typing import Dict, List, Union
 
 from kazoo.client import KazooClient
 from kazoo.exceptions import NodeExistsError, NoNodeError
 from kazoo.handlers.threading import KazooTimeoutError
 from kazoo.retry import KazooRetry
-
-from openeo_driver.backend import UserDefinedProcessMetadata, UserDefinedProcesses
+from openeo_driver.backend import UserDefinedProcesses, UserDefinedProcessMetadata
 from openeo_driver.errors import ProcessGraphNotFoundException
+
 from openeogeotrellis.configparams import ConfigParams
 
 
@@ -20,7 +19,7 @@ class ZooKeeperUserDefinedProcessRepository(UserDefinedProcesses):
 
     _log = logging.getLogger(__name__)
 
-    def __init__(self, hosts: List[str], root: str = "/openeo/udps", zk_client_reuse: bool = False):
+    def __init__(self, hosts: List[str], root: str = "/openeo/udps", *, zk_client_reuse: bool = False):
         self._hosts = ",".join(hosts)
         self._root = root
         self._zk_client_reuse = zk_client_reuse
@@ -28,13 +27,11 @@ class ZooKeeperUserDefinedProcessRepository(UserDefinedProcesses):
 
     @staticmethod
     def _serialize(spec: dict) -> bytes:
-        return json.dumps({
-            'specification': spec
-        }).encode()
+        return json.dumps({"specification": spec}).encode("utf8")
 
     @staticmethod
     def _deserialize(data: bytes) -> dict:
-        return json.loads(data.decode())
+        return json.loads(data.decode("utf8"))
 
     def save(self, user_id: str, process_id: str, spec: dict) -> None:
         with self._zk_client() as zk:
@@ -141,7 +138,7 @@ class InMemoryUserDefinedProcessRepository(UserDefinedProcesses):
 
 
 def main():
-    repo = ZooKeeperUserDefinedProcessRepository(hosts=ConfigParams().zookeepernodes)
+    repo = ZooKeeperUserDefinedProcessRepository(hosts=ConfigParams().zookeepernodes, zk_client_reuse=True)
 
     user_id = 'vdboschj'
     process_graph_id = 'evi'
--- a/tests/test_user_defined_process_repository.py
+++ b/tests/test_user_defined_process_repository.py
@@ -1,5 +1,6 @@
 import collections
 import json
+from typing import Iterator
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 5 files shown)


soxofaan:
Still to do if this would be useful: expose with config and add some tests

openeogeotrellis/user_defined_process_repository.py
```diff
@@ -82,16 +84,32 @@ def delete(self, user_id: str, process_id: str) -> None:
 
     @contextlib.contextmanager
     def _zk_client(self):
-        kz_retry = KazooRetry(max_tries=10, delay=0.5, backoff=2)
-        zk = KazooClient(hosts=self._hosts,connection_retry=kz_retry,
-                 command_retry=kz_retry, timeout=3.0)
-        zk.start(timeout=15.0)
+        if not self._zk_client_reuse or self._zk_client_cache is None:
```
pvbouwel, line 87: Since `self._zk_client_cache is None` is part of the new initialization I would limit the check tot that as it is more readable imho
```suggestion
        if not self._zk_client_cache is None:
```

pvbouwel: If it is to protect against programming errors I would not do it in the same if branch but rather split it out and assert:
```
if not self._zk_client_reuse:
  assert self._zk_client_cache is None, "cache is only for client reuse so must be None if re-use is disabled"
```

pvbouwel (commented):
Just a nit on the if condition but looks good to me.

openeogeotrellis/user_defined_process_repository.py
soxofaan: I indeed considered initially what you suggested there, but I don't completely agree about the readability. 
It's true that combining `not` and `or` in a condition can be a bit tricky, but I kind of like what the current condition clearly expresses: create a new client in these two situations:
- reuse is disabled
- reuse is enabled, but it's the first time we're here

By removing the mention of `zk_client_reuse` in the condition, you obfuscate how that reasoning and you have to understand how `_zk_client_cache` is handled to fully grasp that.

Would it be better to flip the argument `_zk_client_reuse` to e.g. `_zk_recreate_client` so that the condition becomes (without a `not`):

```
if  self._zk_recreate_client or self._zk_client_cache is None:
```

another trick is introducing a new variable to clarify the condition:
```
need_to_create_new_client = not self._zk_client_reuse or self._zk_client_cache is None
if need_to_create_new_client:
```

pvbouwel: Either suggestion is an improvement and both would be OK to me. 
I might lean slightly to the 2nd but I cannot express clearly why so either would work for me

openeogeotrellis/user_defined_process_repository.py
```diff
@@ -20,9 +20,11 @@ class ZooKeeperUserDefinedProcessRepository(UserDefinedProcesses):
 
     _log = logging.getLogger(__name__)
 
-    def __init__(self, hosts: List[str], root: str = "/openeo/udps"):
-        self._hosts = ','.join(hosts)
+    def __init__(self, hosts: List[str], root: str = "/openeo/udps", zk_client_reuse: bool = False):
```
soxofaan, line 23: TODO: config to set this `zk_client_reuse` flag

soxofaan (commented):
add some tests too

openeogeotrellis/user_defined_process_repository.py
soxofaan: Make `zk_client_reuse` keyword-only arg as well

tests/test_user_defined_process_repository.py
```diff
@@ -114,3 +126,39 @@ def test_delete_miss(self, udp_db, zk):
         zk.delete.side_effect = NoNodeError()
         with pytest.raises(ProcessGraphNotFoundException):
             udp_db.delete(user_id="john", process_id="evi1")
+
+    @pytest.mark.parametrize(
+        ["zk_client_reuse", "expected_create_call_counts", "expected_cleanups"],
```
pvbouwel, line 131: I like that it tracks the evolution.

nit: `expected_create_call_counts` and `expected_cleanups`  are both tracking call counts so both parameters should be named similarly so perhaps have expected_cleanups als be expected_cleanup_call_counts

pvbouwel (approved):
Nothing of concern just 2 minor remarks.

tests/test_user_defined_process_repository.py
```diff
@@ -73,10 +73,23 @@ def udp_db(self) -> ZooKeeperUserDefinedProcessRepository:
         return udp_db
 
     @pytest.fixture
-    def zk(self) -> mock.MagicMock:
+    def kazoo_client_factory(self) -> Iterator[mock.MagicMock]:
```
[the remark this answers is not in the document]
soxofaan: as discussed irl: a simple `Iterator` is just fine nowadays

tests/test_user_defined_process_repository.py
soxofaan: done

tests/test_user_defined_process_repository.py
```diff
@@ -73,10 +73,23 @@ def udp_db(self) -> ZooKeeperUserDefinedProcessRepository:
...
+            yield KazooClient
+
+    @pytest.fixture
+    def zk(self, kazoo_client_factory) -> mock.MagicMock:
         """Fixture for `KazooClient instance used by ZooKeeperUserDefinedProcessRepository """
-        with mock.patch('openeogeotrellis.user_defined_process_repository.KazooClient') as KazooClient:
-            yield KazooClient()
+        return kazoo_client_factory()
+
+    def _build_get_return_value(self, *, process_id="evi", pg=PG_2PLUS3):
```
soxofaan, line 89: document this helper

Issue #1043, opened 2025-02-06
load_stac: optimize resolution


bossie:
`j-25020610185143a6a705ec2701413e4e` by @filipsch only requests 60m bands but result has a 10m resolution.

Process graph:
```json
{
  "process_graph": {
    "loadstac1": {
      "arguments": {
        "bands": [
          "B01_60m",
          "B02_60m",
          "B03_60m",
          "B04_60m",
          "B8A_60m",
          "B09_60m",
          "B11_60m",
          "B12_60m",
          "SCL_60m"
        ],
        "featureflags": {
          "indexreduction": 3,
          "tilesize": 256,
          "use-filter-extension": "cql2-json"
        },
        "properties": {
          "eo:cloud_cover": {
            "process_graph": {
              "lte1": {
                "arguments": {
                  "x": {
                    "from_parameter": "value"
                  },
                  "y": 90
                },
                "process_id": "lte",
                "result": true
              }
            }
          },
          "grid:code": {
            "process_graph": {
              "eq2": {
                "arguments": {
                  "x": {
                    "from_parameter": "value"
                  },
                  "y": "MGRS-36NYH"
                },
                "process_id": "eq",
                "result": true
              }
            }
          },
          "processing:version": {
            "process_graph": {
              "eq1": {
                "arguments": {
                  "x": {
                    "from_parameter": "value"
                  },
                  "y": "05.00"
                },
                "process_id": "eq",
                "result": true
              }
            }
          }
        },
        "spatial_extent": {
          "crs": "EPSG:32636",
          "east": 809760,
          "north": 300000,
          "south": 190200,
          "west": 699960
        },
        "temporal_extent": [
          "2020-03-22T00:00:00Z",
          "2020-03-23T00:00:00Z"
        ],
        "url": "https://stac.dataspace.copernicus.eu/v1/collections/sentinel-2-l2a"
      },
      "process_id": "load_stac"
    },
    "saveresult1": {
      "arguments": {
        "data": {
          "from_node": "loadstac1"
        },
        "format": "GTiff",
        "options": {
          "file_metadata": {
            "product_grid": "MGRS tiling grid",
            "product_tile": "36NYH"
          },
          "filename_prefix": "LCFM_CLOUD_DETECTION_TEST_CDSE_STAC_2020_36NYH",
          "separate_asset_per_band": false
        }
      },
      "process_id": "save_result",
      "result": true
    }
  }
}
```

bossie:
[...]

PR #1044, opened 2025-02-06
support "proj:code" in addition to "proj:epsg"

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1043


openeogeotrellis/load_stac.py
```diff
@@ -410,11 +428,18 @@ def intersects_temporally(interval) -> bool:
...
         for asset_id, asset in band_assets.items():
             asset_band_names = get_band_names(itm, asset) or [asset_id]
+            proj_epsg, proj_bbox, proj_shape = get_proj_metadata(itm, asset)
+
             for asset_band_name in asset_band_names:
                 if asset_band_name not in band_names:
                     band_names.append(asset_band_name)
 
-            proj_epsg, proj_bbox, proj_shape = get_proj_metadata(itm, asset)
+                # TODO: assumes square pixels (probably elsewhere too)
```
bossie, line 437: Address this TODO.

openeogeotrellis/load_stac.py
```diff
@@ -515,6 +509,68 @@ def intersects_temporally(interval) -> bool:
 
     band_names = metadata.band_names
 
+    if apply_lcfm_improvements or env.get(EVAL_ENV_KEY.LOAD_STAC_APPLY_LCFM_IMPROVEMENTS, False):
+        logger.info("applying LCFM resolution improvements")
+
+        requested_band_epgs = [epsgs for n, epsgs in band_epsgs.items() if n in band_names]
```
bossie, line 515: Fix typo in `requested_band_epgs`; improve identifiers altogether.

openeogeotrellis/load_stac.py
```diff
@@ -515,6 +509,68 @@ def intersects_temporally(interval) -> bool:
...
+        unique_epsgs = {epsg for epsgs in requested_band_epgs for epsg in epsgs}
+        requested_band_resolutions = [res for n, res in band_resolution.items() if n in band_names]
+
+        if len(unique_epsgs) == 1 and requested_band_resolutions:  # exact resolution
+            target_epsg = unique_epsgs.pop()
+            cell_width = cell_height = max(requested_band_resolutions)
```
bossie, line 521: Is cell size rather than resolution so should be `min` instead ( ~ `layercatalog.py`); rename `band_resolution` and similars to `band_cell_size` etc?

tests/test_load_stac.py
```diff
@@ -111,14 +111,76 @@ def test_dimensions(urllib_and_request_mock, requests_mock):
...
+        yield jvm_mock
+
+
+@pytest.mark.parametrize(
+    ["band_names", "resolution"],
+    [
+        (["AOT_10m"], 10.0),
+        (["WVP_20m"], 20.0),
+        (["WVP_60m"], 60.0),
+    ],
```
bossie, line 146: Test combination so highest resolution (smallest cell size) gets picked (see above).

openeogeotrellis/load_stac.py
```diff
@@ -515,6 +509,68 @@ def intersects_temporally(interval) -> bool:
 
     band_names = metadata.band_names
 
+    if apply_lcfm_improvements or env.get(EVAL_ENV_KEY.LOAD_STAC_APPLY_LCFM_IMPROVEMENTS, False):
```
bossie, line 512: Move `env` check to call from `layercatalog.py`.

bossie:
Will use resolution and CRS of requested bands where:

- resolution is assumed the same across features;
- CRS can be different across features; if it is the same: use this CRS; if not: default to best UTM (like current implementation).

Set `load_stac_apply_lcfm_improvements` in layer catalog or job_option to enable.

PR #1075, opened 2025-02-28
Move import that might give confusing error even when not used. Should save 40mb RAM too.

EmileSonneveld:
Got this error in a graph with only sentinel2, but with limited python-memory:
```bash
Traceback (most recent call last):
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/worker.py", line 1227, in main
    func, profiler, deserializer, serializer = read_command(pickleSer, infile)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/worker.py", line 90, in read_command
    command = serializer._read_with_length(file)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/serializers.py", line 174, in _read_with_length
    return self.loads(obj)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/serializers.py", line 472, in loads
    return cloudpickle.loads(obj, encoding=encoding)
  File "/opt/openeo/lib/python3.8/site-packages/openeogeotrellis/collections/sentinel3.py", line 19, in <module>
    from scipy.spatial import cKDTree  # used for tuning the griddata interpolation settings
  File "/opt/openeo/lib/python3.8/site-packages/scipy/spatial/__init__.py", line 104, in <module>
    from ._qhull import *
ImportError: libopenblasp-r0-8b9e111f.3.17.so: failed to map segment from shared object
```

### DIFF
--- a/openeogeotrellis/layercatalog.py
+++ b/openeogeotrellis/layercatalog.py
@@ -692,6 +692,7 @@ class GeoPySparkLayerCatalog(CollectionCatalog):
                 float(metadata.get("cube:dimensions", "x", "step")),
                 float(metadata.get("cube:dimensions", "y", "step"))
             )
+            # Only import when sentinel3 layer is used to save ~40Mb RAM:
             from openeogeotrellis.collections import sentinel3
 
             pyramid = sentinel3.pyramid(metadata_properties(),
--- a/openeogeotrellis/layercatalog.py
+++ b/openeogeotrellis/layercatalog.py
@@ -37,7 +37,6 @@ from shapely.geometry.base import BaseGeometry
 from openeogeotrellis import sentinel_hub, datacube_parameters
 from openeogeotrellis.catalogs.creo import CreoCatalogClient
 from openeogeotrellis.collections.s1backscatter_orfeo import get_implementation as get_s1_backscatter_orfeo
-from openeogeotrellis.collections import sentinel3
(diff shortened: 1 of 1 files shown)


soxofaan:
I would put a comment before the local import on why it's a local import now. 
Otherwise it's too tempting to make it top-level import again

EmileSonneveld:
Btw, the command used to measure memory:
```
/usr/bin/time -v python3 -c "from scipy.spatial import cKDTree"
Maximum resident set size (kbytes): 48712
```
Which says 48Mb, but probably scipy gets imported anyway:
```
/usr/bin/time -v python3 -c "import scipy"
Maximum resident set size (kbytes): 30776 
```
But yet again there is more that gets imported in `sentinel3.py`, so the memory guess might not be so accurate as I thought

soxofaan:
> But yet again there is more that gets imported in sentinel3.py,

Indeed, in that sentinel3.py, I don't see a lot of special imports. Most things are typically already imported anyway, except for that `from scipy.spatial import cKDTree` . And as you illustrated above, that just 18MB. 
You could also make the `scipy.spatial.cKDTree` import local, as it only used in one place at the moment

EmileSonneveld:
The main point is here to remove the misplaces sentinel3 error, the performance issue is a bonus.
Using a local import makes the code not "fail fast", but that should be covered in the integration tests.

soxofaan (approved):

Issue #1073, opened 2025-02-27
StacApiWorkspace: retry create collection/item


bossie:
`j-25022709071249edbd170df80a421bd9` failed because of a transient error while uploading a STAC item:

```
Traceback (most recent call last):
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 716, in urlopen
    httplib_response = self._make_request(
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 468, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 463, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib64/python3.8/http/client.py", line 1347, in getresponse
    response.begin()
  File "/usr/lib64/python3.8/http/client.py", line 307, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python3.8/http/client.py", line 268, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib64/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib64/python3.8/ssl.py", line 1241, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/lib64/python3.8/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 104] Connection reset by peer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/openeo/lib/python3.8/site-packages/requests/adapters.py", line 667, in send
    resp = conn.urlopen(
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 802, in urlopen
    retries = retries.increment(
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/util/retry.py", line 552, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/packages/six.py", line 769, in reraise
    raise value.with_traceback(tb)
  File "/opt/openeo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 716, in urlopen
    httplib_response = self._make_request(
[...]

PR #1080, opened 2025-02-28
retry creation of STAC collection/item

bossie:
#1073


bossie:
`openeogeotrellis.integrations.stac.StacApiIO` (and its example `pystac.stac_io.RetryStacIO`) mix `urllib` and `urllib3` APIs with regards to handling HTTP errors so replaced with `openeogeotrellis.integrations.stac.ResilientStacIO` based on `requests`. This allows for better mocking of HTTP calls as well.

`requests_mock` [does not allow for testing retries](https://github.com/jamielennox/requests-mock/issues/83) so using `responses` instead (a maintained version of `httpretty`).

Issue #846, opened 2024-08-22
support adding temporal dimension


jdries:
it is possible to add a time dimension, but the rdd remains the same (spatial).
It should be quite easy to convert the keys and metadata approperiately. This will fix a number of cases where users do not get the expected behaviour after adding temporal dim.

JeroenVerstraelen:
This could be used in GeopysparkDataCube.add_dimension() with the following (pseudo) code:
https://gist.github.com/JeroenVerstraelen/99b19cca97a6beed7470866999706666

PR #1087, opened 2025-03-05
issue 846 change temporal layer in add_dimension

ElienVandermaesenVITO:
#846

### DIFF
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -362,9 +362,22 @@ class GeopysparkDataCube(DriverDataCube):
 
     @callsite
     def add_dimension(self, name: str, label: str, type: str = None):
+        cube = self
+        if type== "temporal":
+            if cube.metadata.has_temporal_dimension():
+                raise OpenEOApiException(status_code=400, code="DimensionAlreadyExist", message=
+                """Temporal dimension can only be added if temporal dimension does not exist""")
+            jvm = get_jvm()
+            label_as_zoneddatetime = jvm.java.time.ZonedDateTime.apply()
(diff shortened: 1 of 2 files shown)


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -362,9 +362,22 @@ def _normalize_temporal_reducer(self, dimension: str, reducer: str) -> str:
 
     @callsite
     def add_dimension(self, name: str, label: str, type: str = None):
+        cube = self
+        if type== "temporal":
+            if cube.metadata.has_temporal_dimension():
+                raise OpenEOApiException(status_code=400, code="DimensionAlreadyExist", message=
```
jdries, line 368: Better: log an error message if this happens, then follow up in kibana logs if this occurs in practice.
Then a next step can be to convert the error log in an exception? 
(Unless the current version would also result in an exception already.)

openeogeotrellis/geopysparkdatacube.py
```diff
@@ -362,9 +362,22 @@ def _normalize_temporal_reducer(self, dimension: str, reducer: str) -> str:
...
+                """Temporal dimension can only be added if temporal dimension does not exist""")
+            jvm = get_jvm()
+            label_as_zoneddatetime = jvm.java.time.ZonedDateTime.apply()
+            temporal_key = jvm.geotrellis.layer.TemporalKey.apply(label_as_zoneddatetime)
+
+            def to_temporal_layer(rdd):
```
jdries, line 374: Minor: this could also be a lambda?

jdries (changes requested):
Reviewed, is there also a unit test?

PR #1095, opened 2025-03-12
add pre-commit hook to ensure consistency of version in changelog

soxofaan:


soxofaan:
hmm this approach seems to be challenging to get working in GitKraken (git UI I'm currently using), which fails to find a python executable to execute the script.

maybe doing this check as a unit test is even better as it will also work if developer doesn't have pre-commit enabled

PR #1100, opened 2025-03-14
Do not consult ZK for CONCURRENT_POD_LIMIT if yunikorn_user_specific_queues

tcassaert:
The limit is set on queue level when we're using `yunikorn_user_specific_queues`. 

We can keep the `CONCURRENT_POD_LIMIT` logic for now to limit the queue length. For that we check on `SUBMITTED` apps rather than on `RUNNING` apps.

The CONCURRENT_POD_LIMIT should be set to a reasonable number in the webapp driver. Queued jobs do not use resources, but might put some stress on Yunikorn/etcd so we better limit it anyways, just not as strict as we used to do with the RUNNING apps.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -2059,18 +2059,19 @@ class GpsBatchJobs(backend.BatchJobs):
 
             # Check concurrent_pod_limit constraints.
             concurrent_pod_limit = ConfigParams().concurrent_pod_limit
-            limitting_state = "RUNNING"
             if not get_backend_config().yunikorn_user_specific_queues:
-              try:
-                  with zk_client(hosts=ConfigParams().zookeepernodes) as zk:
-                      zk_path = f"{get_backend_config().zookeeper_root_path}/config/users/{user_id}/concurrent_pod_limit"
-                      concurrent_pod_limit = int(zk.get(zk_path)[0])
-                      log.info(f"Concurrent job limit for user {user_id} found: {concurrent_pod_limit}")
-              except kazoo.exceptions.NoNodeError:
-                  pass
-              except Exception:
-                  log.warning(f"Failed to get user specific concurrent_pod_limit", exc_info=True)
-              limitting_state = "SUBMITTED"
+                try:
+                    with zk_client(hosts=ConfigParams().zookeepernodes) as zk:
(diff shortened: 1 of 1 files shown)


openeogeotrellis/backend.py
```diff
@@ -2031,15 +2031,19 @@ def as_arg_element(dependency: dict) -> dict:
...
-                    zk_path = f"{get_backend_config().zookeeper_root_path}/config/users/{user_id}/concurrent_pod_limit"
-                    concurrent_pod_limit = int(zk.get(zk_path)[0])
-                    log.info(f"Concurrent job limit for user {user_id} found: {concurrent_pod_limit}")
-            except kazoo.exceptions.NoNodeError:
-                pass
-            except Exception:
-                log.warning(f"Failed to get user specific concurrent_pod_limit", exc_info=True)
+            limitting_state = "RUNNING"
+            if not get_backend_config().yunikorn_user_specific_queues:
+              try:
```
soxofaan, line 2036: use 4 spaces indentation. Initially I didn't even noticed that there was added indentation

openeogeotrellis/backend.py
```diff
@@ -2031,15 +2031,19 @@ def as_arg_element(dependency: dict) -> dict:
...
+                  with zk_client(hosts=ConfigParams().zookeepernodes) as zk:
+                      zk_path = f"{get_backend_config().zookeeper_root_path}/config/users/{user_id}/concurrent_pod_limit"
+                      concurrent_pod_limit = int(zk.get(zk_path)[0])
+                      log.info(f"Concurrent job limit for user {user_id} found: {concurrent_pod_limit}")
+              except kazoo.exceptions.NoNodeError:
+                  pass
+              except Exception:
+                  log.warning(f"Failed to get user specific concurrent_pod_limit", exc_info=True)
+              limitting_state = "SUBMITTED"
```
soxofaan, line 2045: I'm a bit confused here, should't the construct be somehting like

```python
if not get_backend_config().yunikorn_user_specific_queues:
    ...
    limitting_state = "RUNNING"
else:
    limitting_state = "SUBMITTED"
```

to get the original behavior.

Now it's the other way around unless I'm misunderstanding

PR #1112, opened 2025-03-25
Make urllib.error.HTTPError log the URL with the issue.

EmileSonneveld:
No ticket. Just found many errors like this on Kibana: `urllib.error.HTTPError: HTTP Error 404: Not Found`

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import time
 import traceback
+import urllib
 import uuid
 import importlib.metadata
 from copy import deepcopy
@@ -1166,7 +1167,6 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
             summary = str_truncate(summary, width=width)
         else:
             is_client_error = False  # Give user the benefit of doubt.
-            import urllib
             if isinstance(error, FileNotFoundError):
                 summary = repr_truncate(str(error), width=width)
             elif isinstance(error, urllib.error.HTTPError):
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1166,8 +1166,11 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
             summary = str_truncate(summary, width=width)
         else:
             is_client_error = False  # Give user the benefit of doubt.
+            import urllib
             if isinstance(error, FileNotFoundError):
                 summary = repr_truncate(str(error), width=width)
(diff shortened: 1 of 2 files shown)


bossie:
Do you have a Kibana link?

EmileSonneveld:
[yes (internal link tough)](https://kibana-infra.vgt.vito.be/app/kibana#/discover/9385d820-3cdf-11ee-8cc4-3747d5233c59?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2d,to:now))&_a=(columns:!(message,kubernetes.pod_name,filename,user_id,kubernetes.pod_labels.user),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'07fb0200-6989-11ed-8cc4-3747d5233c59',key:levelname,negate:!f,params:(query:ERROR),type:phrase),query:(match:(levelname:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'07fb0200-6989-11ed-8cc4-3747d5233c59',key:filename,negate:!f,params:!(views.py,batch_job.py),type:phrases,value:'views.py,%20batch_job.py'),query:(bool:(minimum_should_match:1,should:!((match_phrase:(filename:views.py)),(match_phrase:(filename:batch_job.py))))))),index:'07fb0200-6989-11ed-8cc4-3747d5233c59',interval:auto,query:(language:kuery,query:'message:*HTTPError*'),sort:!(!('@timestamp',desc))))

openeogeotrellis/backend.py
```diff
@@ -1166,8 +1166,11 @@ def get_missing_sentinel1_band() -> Optional[str]:
             summary = str_truncate(summary, width=width)
         else:
             is_client_error = False  # Give user the benefit of doubt.
+            import urllib
```
bossie, line 1169: Is there a particular reason why you're doing a function-level `import` rather than a top-level `import`?

EmileSonneveld: I thought it would save memory, but it it would be a few kb tops.

EmileSonneveld:
On CDSE staging, this will now show errors like this to the python client:
`openeo.rest.OpenEoApiError: [500] Internal: Server error: 'HTTP Error 404: Not Found: https://artifactory.vgt.vito.be:443/artifactory/auxdata-public/hrlvlcc/pardel/PARDEL_2022_E512N276-03035-010m_V100_CT_masked.geojson' (ref: r-25032808235849b58572b511c3e5af32)`

There are some places, like in validate where `summarize_exception` is not used yet, giving it the error without context.

PR #1115, opened 2025-03-27
config: mandate the region for workspaces

pvbouwel:
Initially the workspace region parameter was optional in order to be able to deploy without breaking existing configs.

The region information is however crucial for correct workspace operation so this change makes it mandatory.


bossie (approved):
LGTM

Issue #1069, opened 2025-02-25
sentinel3: memory allocation error


jdries:
A simple Sentinel-3 reading goes OOM while reading coordinates.
Perhaps  this requires lazy loading via dask, as we usually don't need the whole dataset.

j-2502250813404ad18aa7cd87ee428aaa

[...]

PR #1118, opened 2025-03-28
disable decode_cf on open_dataset to save memory for sentinel3 #1069

JeroenVerstraelen:
`lat_lon_ds = xr.open_dataset(latlon_file, drop_variables=to_drop, decode_cf=False)`
decode_cf rescales the original netcdf data from int32 to float64 causing OOM errors on the spark executors. By keeping the data as int32 and scaling the bbox instead we can significantly lower memory usage.

Note that this requires an update of `tests/data/binary/sentinel3_edge_ref.tif` reference data. Because we can now work with float64 for lat_orig and lon_orig as these are smaller arrays. Previously we cast everything to float32 while the precision of these NetCDFs is 1e-06 which meant the last digit(s) in some coordinates were rounded. I verified that I get the exact old reference data by just casting these arrays back to float32:
```
lat_orig = lat_orig.astype(np.float32)
lon_orig = lon_orig.astype(np.float32)
```
If required we can cast these to float32 to avoid slightly changing existing workflows.

I went over the decode_cf implementation and verified that the original NetCDF data is not changed in other ways. I did make the assumption that a FillValue will never be required in our usecase, even though the NetCDF does declare a FillValue of -2147483648.

### DIFF
--- a/openeogeotrellis/collections/sentinel3.py
+++ b/openeogeotrellis/collections/sentinel3.py
@@ -572,9 +572,9 @@ def _read_latlonfile(bbox, latlon_file, lat_band="latitude", lon_band="longitude
     lon_offset = lat_lon_ds[lon_band].attrs.get('add_offset', 0.0)
 
     # lat_lon_ds is scaled, so we need to inverse scale the bbox as well.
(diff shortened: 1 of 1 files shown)


jdries (approved):
Great find! Makes a whole lot of sense.
I assume this is covered by existing tests?

EmileSonneveld:
Nice, how much memory is saved?

JeroenVerstraelen:
> Nice, how much memory is saved?

Judging from tests on the staging cluster we can run the job with `"executor-memoryOverhead": "1800m"` now. Not sure if the default options on staging have an effect but on prod it required a memoryOverhead of 2800m before this optimization.

PR #1119, opened 2025-03-28
feature: Allow S3 presigned urls against S3 proxy

pvbouwel:
If the geopyspark-driver has config set to be able to issue IDP tokens and the STS endpoint as well as a new config `s3_region_proxy_endpoints` which has the S3 proxy endpoint per region then it will generate presigned S3 urls for assets as long as it knows about the asset buckets via a workspace.

### DIFF
--- a/openeogeotrellis/asset_urls.py
+++ b/openeogeotrellis/asset_urls.py
@@ -1,0 +1,41 @@
+import logging
+from typing import Tuple
+from urllib.parse import urlparse
+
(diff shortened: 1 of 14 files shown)


pvbouwel:
This requires https://github.com/Open-EO/openeo-python-driver/pull/387 (which will cause build failures)
The changes in conftest allows for easier overrides and thus as a workaround for  https://github.com/Open-EO/openeo-python-driver/issues/265

PR #1131, opened 2025-04-03
Attempt to avoid docker.io Server message: toomanyrequests

EmileSonneveld:
Attempt to avoid rate limiting
Made ticket for this: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1132

### DIFF
--- a/openeogeotrellis/integrations/calrissian.py
+++ b/openeogeotrellis/integrations/calrissian.py
@@ -199,7 +199,7 @@ class CalrissianJobLauncher:
 
         container = kubernetes.client.V1Container(
             name="calrissian-input-staging",
-            image="alpine:3",
+            image="docker-upstream.artifactory.vgt.vito.be/alpine:3.17.0",
             security_context=self._security_context,
             command=["/bin/sh"],
             args=["-c", f"set -euxo pipefail; echo '{cwl_serialized}' | base64 -d > {cwl_path}"],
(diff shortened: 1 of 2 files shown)


openeogeotrellis/integrations/calrissian.py
```diff
@@ -199,7 +199,7 @@ def create_input_staging_job_manifest(self, cwl_source: CwLSource) -> Tuple[kube
 
         container = kubernetes.client.V1Container(
             name="calrissian-input-staging",
-            image="alpine:3",
+            image="docker-upstream.artifactory.vgt.vito.be/alpine:3.17.0",
```
soxofaan, line 202: I think we should avoid putting `artifactory.vgt.vito.be`  references in openeo-geopyspark-driver code

EmileSonneveld: Would something like docker://registry.stag.warsaw.openeo.dataspace.copernicus.eu/rand/alpine:3 be better?

soxofaan: Ideally, we just have an "abstract" `alpine:3` reference here, but there is a way to configure the underlying k8s machinery to resolve this from the desired docker repository (e.g. like you can have global `pip` config files to set a prefered "index-url" instead of the default pypi.org)


Alternatively, we introduce a config field here so it can be override by a deployment

EmileSonneveld:
FFR `registry.stag.warsaw.openeo.dataspace.copernicus.eu/rand/alpine:3` was manually uploaded with skopeo:
`skopeo copy --multi-arch=all --format=oci docker-daemon:alpine:3 docker://registry.stag.warsaw.openeo.dataspace.copernicus.eu/rand/alpine:3`
This is not ideal, better there would be some caching between the pods and docker.io or anything else

PR #1147, opened 2025-04-16
1127 load stac produces wrong epsg

JeroenVerstraelen:
Changes to geopyspark (0.4.9+openeo):
https://github.com/Open-EO/geopyspark/commits/geotrellis_3.5

### DIFF
--- a/openeogeotrellis/geopysparkdatacube.py
+++ b/openeogeotrellis/geopysparkdatacube.py
@@ -2404,14 +2404,14 @@ class GeopysparkDataCube(DriverDataCube):
             result = xr.DataArray(collection[0][1], dims=dims, coords=coords)
 
         # add some metadata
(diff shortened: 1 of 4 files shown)


openeogeotrellis/geopysparkdatacube.py
```diff
@@ -1482,7 +1482,8 @@ def linear_scale_range(self, input_min, input_max, output_min, output_max) -> 'G
     def timeseries(self, x, y, srs="EPSG:4326") -> Dict:
         # TODO #421 drop old unsued "point timeseries" feature
         max_level = self.get_max_level()
```
jdries, line 1484: is this method still used? Maybe deleting it is better to reduce technical debt?

jdries (approved):
approved, do note that these two fixes are in 'dead' code paths. I'm wondering if we could avoid future maintenance effort by simply deleting them.
Also because more code is perhaps harder to grasp for new developers?

PR #1154, opened 2025-04-22
issue 384 netcdf add scale and offset bands metadata

ElienVandermaesenVITO:
issue 384 https://github.com/Open-EO/openeo-geotrellis-extensions/issues/386


jdries:
I'll have to merge this one to fix build.
A 'NetCDFOptions' object would have been better to avoid having to always adapt the API.

PR #1157, opened 2025-04-23
Feature/s3 assets via s3 proxy

pvbouwel:
If the geopyspark-driver has config set to be able to issue IDP tokens and the STS endpoint as well as a new config s3_region_proxy_endpoints which has the S3 proxy endpoint per region then it will generate presigned S3 urls for assets as long as it knows about the asset buckets via a workspace.

This builds on top of https://github.com/Open-EO/openeo-python-driver/pull/387
The changes in conftest allows for easier overrides and thus as a workaround for https://github.com/Open-EO/openeo-python-driver/issues/265

This replaces #1119

### DIFF
--- a/openeogeotrellis/asset_urls.py
+++ b/openeogeotrellis/asset_urls.py
@@ -1,0 +1,41 @@
+import logging
+from typing import Tuple
+from urllib.parse import urlparse
+
+from openeo_driver.asset_urls import AssetUrl
+from openeogeotrellis.integrations.s3proxy.exceptions import ProxyException
+from openeogeotrellis.integrations.s3proxy.s3 import get_proxy_s3_client_for_job
+from openeogeotrellis.integrations.s3 import create_presigned_url
+
+_log = logging.getLogger(__name__)
+
+
+class PresignedS3AssetUrls(AssetUrl):
+    def __init__(self, expiration: int = 24 * 3600):
+        self._expiration = expiration
+
+    def get(self, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
+        href = asset_metadata.get("href")
+        if isinstance(href, str) and href.startswith("s3://"):
+            try:
+                bucket, key = self.get_bucket_key_from_uri(href)
+                return self._get_presigned_url_against_proxy(bucket, key, job_id, user_id)
+            except (ValueError, ProxyException) as e:
+                logging.debug(f"Falling back to default asset getter because: {e}")
+        return super().get(asset_metadata, asset_name, job_id, user_id)
+
+    @staticmethod
+    def get_bucket_key_from_uri(s3_uri: str) -> Tuple[str, str]:
+        _parsed = urlparse(s3_uri, allow_fragments=False)
+        if _parsed.scheme != "s3":
+            raise ValueError(f"Input {s3_uri} is not a valid S3 URI should be of form s3://<bucket>/<key>")
+        bucket = _parsed.netloc
+        if _parsed.query:
+            key = _parsed.path.lstrip('/') + '?' + _parsed.query
+        else:
+            key = _parsed.path.lstrip('/')
+        return bucket, key
+
+    def _get_presigned_url_against_proxy(self, bucket: str, key: str, job_id: str, user_id: str) -> str:
+        s3_client = get_proxy_s3_client_for_job(bucket, job_id, user_id)
+        return create_presigned_url(s3_client, bucket_name=bucket, object_name=key, expiration=self._expiration)
--- a/openeogeotrellis/asset_urls.py
+++ b/openeogeotrellis/asset_urls.py
@@ -14,7 +14,7 @@ class PresignedS3AssetUrls(AssetUrl):
     def __init__(self, expiration: int = 24 * 3600):
         self._expiration = expiration
 
-    def get(self, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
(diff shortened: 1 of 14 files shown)


jdries:
@pvbouwel  looks good enough for me, do you think detailed review is needed? (I'm assuming we need this for object storage issues?

pvbouwel:
It still requires a config change before behavior changes so I think it not to high risk to not have in-depth review.

openeogeotrellis/asset_urls.py
```diff
@@ -0,0 +1,41 @@
+import logging
+from typing import Tuple
+from urllib.parse import urlparse
+
+from openeo_driver.asset_urls import AssetUrl
+from openeogeotrellis.integrations.s3proxy.exceptions import ProxyException
```
soxofaan, line 6: This introduces toplevel module that adds dependencies under `openeogeotrellis.integrations`, but it's nowhere used in openeo-geopyspark-driver at the moment, so it seems that `PresignedS3AssetUrls` could also live under `openeogeotrellis.integrations.s3proxy` (or something like that?

openeogeotrellis/integrations/identity.py
```diff
@@ -77,6 +77,13 @@ def instance(cls) -> IDPTokenIssuer:
         else:
             return cls()
 
+    def _hard_reset(self) -> None:
+        """
+        This is only for testing purpose because different tests can actually require different idp config.
```
soxofaan, line 82: can't this be solved through config override helpers as used in other tests?

openeogeotrellis/integrations/s3proxy/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class ProxyException(RuntimeError):
+    ...
```
soxofaan, line 2: FYI: for exceptions it's more common to use `pass` (which explicitly means: do nothing)  instead of `...`  (which gives off work-in-progress vibes or even "must be implemented by subclasses" )

soxofaan:
I think this one should also get a CHANGELOG entry

tests/conftest.py
```diff
@@ -300,14 +310,64 @@ def job_registry() -> InMemoryJobRegistry:
...
+    for config_key, config_override_value in _config_overrides.items():
+        if hasattr(OpenEoBackendConfig, config_key):
+            openeo_driver_overrides[config_key] = config_override_value
+        elif hasattr(openeogeotrellis.config.GpsBackendConfig, config_key):
+            geopyspark_driver_overrides[config_key] = config_override_value
+        else:
+            raise ValueError(f"config value {config_key} cannot be attributed to a config")
+    with openeo_driver.testing.config_overrides(**openeo_driver_overrides):
+        with openeogeotrellis.testing.gps_config_overrides(**geopyspark_driver_overrides, **openeo_driver_overrides):
+            yield get_backend_config()
```
soxofaan, line 359: This config override system is getting quite complex with lot of layers. I think we should find a way to simplify it

tests/conftest.py
```diff
@@ -414,8 +500,11 @@ def mock_s3_bucket(mock_s3_resource, monkeypatch):
         yield bucket
 
 
-@pytest.fixture
-def moto_server(monkeypatch) -> str:
+moto_server_address: Optional[str] = None
```
soxofaan, line 503: This variable is used as global variable, which can cause a lot of trouble in tests (e.g. debugging of failing tests, undermines parallelization of tests, ...)

isn't there an alternative, based on fixtures?

openeogeotrellis/asset_urls.py
pvbouwel: Will address in a separate PR and link it to https://github.com/Open-EO/openeo-geopyspark-driver/issues/1176 as well as mention you

openeogeotrellis/integrations/identity.py
pvbouwel: This component allows for live-reloading as long as config is valid.

If invalid config is put in place fatals are logged but no actual failure is being done as that could give an availability issue. For testing however it is important to be able to reset to cases where there is no config as well but such a config change won´t make changes because of the behavior detailed above. So I don´t see how to do this with config

openeogeotrellis/integrations/s3proxy/exceptions.py
pvbouwel: Will change remaining `...`

tests/conftest.py
pvbouwel: I am open to suggestions but the problem with fixtures is that this address is only known once the motoserver is started but it needs to be injected into the config when it is started. Since motoserver is not needed for most tests I did not really want to have it as a dependency for all the tests. I guess technically it is possible to have 1 fixture (e.g. enable_moto_server) influence the behavior of another fixture (like the moto_server) fixture such that I can have config depend on the moto_server fixture where it is most of the times lightweight (since enable_moto_server is disabled and it would just yield None). But that introduces the complexity that is pointed out for the override system.

But I wonder whether that is a better place to be in because it makes the usage of moto_server more complicated because using it by itself won´t be enough?

It also won´t solve parallelization when parallellization is done in a single process because the original moto_server fixtures uses env variables to pass SWIFT_URL but in threaded parallellization that will give problems as well.

tests/conftest.py
pvbouwel: I have pushed one more commit to https://github.com/Open-EO/openeo-geopyspark-driver/pull/1206 to remove 1 of the fixtures (inline it) and make it clear which fixtures are internal

Issue #1062, opened 2025-02-20
Geotiff format: document bands_metadata and file_metadata


jdries:
These options do not seem to be described in format options, please add:
https://documentation.dataspace.copernicus.eu/APIs/openEO/File_formats.html

bossie:
FYI coincidentally I did this for `file_metadata` yesterday (concise): https://github.com/Open-EO/openeo-geopyspark-driver/blob/f05ecc006b638529e8bec3b6b598efbe6dac1aa2/openeogeotrellis/backend.py#L515-L518

PR #1163, opened 2025-04-25
issue 1062 add bands_metadata documentation

ElienVandermaesenVITO:
#1062


openeogeotrellis/backend.py
```diff
@@ -521,6 +521,11 @@ def file_formats(self) -> dict:
                             "type": ["object"],
                             "description": "Sets file-level GDAL metadata (key-value).",
                         },
+                        "bands_metadata": {
+                            "type": ["object"],
```
soxofaan, line 525: if it's optional (None by default), the type should reflect that:
```suggestion
                            "type": ["object", "null"],
```

Issue #1172, opened 2025-04-30
StacApiWorkspace: access token expires during export


bossie:
Reported in https://github.com/ESA-WEED-project/eo_processing/issues/130

That's actually this TODO:

https://github.com/Open-EO/openeo-geopyspark-driver/blob/afefd0c8b825cc38253c5fe0a3ffac99c0a3a442/openeogeotrellis/workspace/stac_api_workspace.py#L97-L98

bossie:
Related: https://github.com/eu-cdse/openeo-cdse-infra/issues/496

JeroenVerstraelen:
![Image](https://github.com/user-attachments/assets/dc410dc7-9cc7-403c-993a-f1b646955261)

JeroenVerstraelen:
Verified locally and merged this asap as this is a blocker for the WEED project. As @bossie mentioned, the following tasks are left as TODO:
- get the expiry time from the access token itself (exp property) and set cache TTL to e.g. 75% of this time; this means adapting the fake access token in the unit test;
- use Session with retries;
- preferably, backport the fresh argument and exp property to openeo_driver.util.auth.ClientCredentialsAccessTokenHelper (check with @soxofaan)
- adapt backendconfig_prod.py to make use of openeogeotrellis.workspace.vito_stac_api_workspace (you'll have to because it's a breaking change). This is already the case on dev and staging.

Next step is testing this on dev and staging. If that works we can adjust backendconfig_prod.py and promote.

soxofaan:
> We need to watch out for promotions to prod before backendconfig_prod.py is adjusted because the constructor of the StacApiWorkspace changed. The get_access_token function we pass now requires a fresh parameter, which is not present in the current config.

I think we have to be careful with putting too much experimental logic into config files like `backendconfig_prod.py` as this breaks the flexibility to decouple the config from the implementation. At the moment it's also impossible to unit tests logic that's defined in `backendconfig_prod.py`.

JeroenVerstraelen:
The export_workspace pg for WEED now works on CDSE staging:
https://gist.github.com/JeroenVerstraelen/4578f38b5b67859a4070ae0e6ae32b60

It uses the new WAW4-1 buckets since the ones on WAW3-1 are currently broken.

The only thing left to do is adjust `backendconfig_prod.py` to use the helper function and then promote to prod.

JeroenVerstraelen:
[...]

PR #1173, opened 2025-04-30
handle expired access token

bossie:
#1172

### DIFF
--- a/openeogeotrellis/workspace/helpers.py
+++ b/openeogeotrellis/workspace/helpers.py
@@ -1,9 +1,11 @@
+import time
 from pathlib import PurePath, Path
 from typing import Callable
 from urllib.parse import urlparse
 
 import pystac
-from openeo_driver.util.auth import ClientCredentialsAccessTokenHelper, ClientCredentials
+from openeo_driver.util.auth import _AccessTokenCache
+from openeo.rest.auth.oidc import OidcClientCredentialsAuthenticator, OidcClientInfo, OidcProviderInfo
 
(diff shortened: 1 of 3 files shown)


bossie:
Currently, only one OIDC access token is requested for the entire `merge` call. In this case, the access token has an expiry time of 5 minutes so for large jobs, it will expire before `merge` completes.

This PR allows `merge` to cope with an expired access token by (for each operation on the STAC API):
 1. expecting a 401 response;
 2. requesting a fresh access token;
 3. retrying the original operation once again with this new access token.
 
Points of concern:
 - the access token should still be cached like in the original implementation (= `ClientCredentialsAccessTokenHelper`);
 - requesting an access token should not hinder application startup (https://github.com/eu-cdse/openeo-cdse-infra/issues/496);
 - requesting an access token should be resilient (use `Session` with retries).
 - the signature of `get_access_token` changes (accepts a `fresh` argument) so clients have to adapt.

bossie:
To wrap this up, essentially address the remaining TODOs.

- get the expiry time from the access token itself (`exp` property) and set cache TTL to e.g. 75% of this time; this means adapting the fake access token in the unit test;
- use `Session` with retries;
- preferably, backport the `fresh` argument and `exp` property to `openeo_driver.util.auth.ClientCredentialsAccessTokenHelper` (check with @soxofaan)
- adapt `backendconfig_prod.py` to make use of `openeogeotrellis.workspace.vito_stac_api_workspace` (you'll have to because it's a breaking change). This is already the case on dev and staging.

PR #1188, opened 2025-05-08
Test showing calrissian logs to user on error. 

EmileSonneveld:
@soxofaan Would it be safe to show logs like this in the error message?
It would help to inspect tasks without having access to the VPN
https://github.com/eu-cdse/openeo-cdse-infra/issues/484

### DIFF
--- a/openeogeotrellis/integrations/calrissian.py
+++ b/openeogeotrellis/integrations/calrissian.py
@@ -408,7 +408,22 @@ class CalrissianJobLauncher:
         elif final_status is None:
             raise TimeoutError(f"CWL Job {job_name} did not finish within {timeout}s")
         elif final_status != "complete":
-            raise RuntimeError(f"CWL Job {job_name} failed with {final_status=} after {timer.elapsed()=:.2f}s")
+            # get logs from latest pod in job:
+            pods = kubernetes.client.CoreV1Api().list_namespaced_pod(
+                namespace=self._namespace, label_selector=f"job-name={job_name}", watch=False
+            )
+            last_pod = sorted(pods.items, key=lambda p: p.metadata.creation_timestamp, reverse=True)[0]
+            # get logs from pod:
+            logs = kubernetes.client.CoreV1Api().read_namespaced_pod_log(
+                name=last_pod.metadata.name,
+                namespace=self._namespace,
+                container=last_pod.spec.containers[0].name,
+                follow=False,
+            )
+            logs = "\n".join(logs.splitlines()[-5:])
+            raise RuntimeError(
+                f"CWL Job {job_name} failed with {final_status=} after {timer.elapsed()=:.2f}s and last lines of failed pod: {logs}"
+            )
         else:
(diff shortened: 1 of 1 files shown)


openeogeotrellis/integrations/calrissian.py
```diff
@@ -408,7 +408,22 @@ def launch_job_and_wait(
         elif final_status is None:
             raise TimeoutError(f"CWL Job {job_name} did not finish within {timeout}s")
         elif final_status != "complete":
-            raise RuntimeError(f"CWL Job {job_name} failed with {final_status=} after {timer.elapsed()=:.2f}s")
+            # get logs from latest pod in job:
+            pods = kubernetes.client.CoreV1Api().list_namespaced_pod(
+                namespace=self._namespace, label_selector=f"job-name={job_name}", watch=False
+            )
+            last_pod = sorted(pods.items, key=lambda p: p.metadata.creation_timestamp, reverse=True)[0]
```
soxofaan, line 415: don't `sort` to just get the last one, use `max`:
```suggestion
            last_pod = max(pods.items, key=lambda p: p.metadata.creation_timestamp)
```

openeogeotrellis/integrations/calrissian.py
```diff
@@ -408,7 +408,22 @@ def launch_job_and_wait(
...
+            # get logs from pod:
+            logs = kubernetes.client.CoreV1Api().read_namespaced_pod_log(
+                name=last_pod.metadata.name,
+                namespace=self._namespace,
+                container=last_pod.spec.containers[0].name,
+                follow=False,
+            )
+            logs = "\n".join(logs.splitlines()[-5:])
```
soxofaan, line 423: I would not `"\n".join`  multiple lines, an exception message is not a multiline text context,

soxofaan (commented):
yes interesting idea

how do these logs typically look in this case. I mean, are the lasts 5 lines the most relevant ones usually?


Also: I would put a try-except around the attempt to get the logs to avoid that some random failure to get the logs completely obfuscates what actually went wrong for the user

EmileSonneveld:
An example of an error with a mount:
`RuntimeError('CWL Job r-2505081502114ba7b7-cal-cwl-1005d002 failed with final_status=\'failed\' after timer.elapsed()=20.08s and last lines of failed pod:   File "/usr/local/lib/python3.10/shutil.py", line 374, in copystat\n    lookup("utime")(dst, ns=(st.st_atime_ns, st.st_mtime_ns),\nPermissionError: [Errno 1] Operation not permitted\nStarting Cleanup\nFinishing Cleanup') (ref: r-2505081502114ba7b79ae8d0541fd914)`

openeogeotrellis/integrations/calrissian.py
EmileSonneveld: at the moment there are no clean error messages. Once I get calrissian working again I could check if sommarize_exception would work here

PR #1190, opened 2025-05-09
workspaces: actually use region specific config

pvbouwel:
Workspaces track a region but client config just takes environment variable which assume workspace is in the same region as the execution environment its object storage.

This change adds openeogeotrellis.integrations.s3.get_s3_client which takes a region and then configures the client as required.

### DIFF
--- a/openeogeotrellis/integrations/s3/__init__.py
+++ b/openeogeotrellis/integrations/s3/__init__.py
@@ -1,2 +1,3 @@
 from openeogeotrellis.integrations.s3.bucket_details import BucketDetails, is_workspace_bucket
 from openeogeotrellis.integrations.s3.presigned_url import create_presigned_url
+from openeogeotrellis.integrations.s3.client import get_s3_client
--- a/openeogeotrellis/integrations/s3/client.py
+++ b/openeogeotrellis/integrations/s3/client.py
@@ -1,0 +1,16 @@
+from __future__ import annotations
+import boto3
+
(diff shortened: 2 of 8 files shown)


soxofaan:
quick observation: any reason you call this new module `openeogeotrellis.integrations.s3_client` instead of just `openeogeotrellis.integrations.s3`? 


I also see a lot of CDSE/CloudFerro references which we usually want to eliminate from the generic openeo-geopyspark-driver project

pvbouwel:
@soxofaan 

s3 depends on workspace Object storage but workspace depends on being able to get a generic s3 client so if I put this code under s3 then I have a dependency cycle. But I am as well contemplating whether having it all under s3 package and just have a local import for `ObjectStorageWorkspace` in bucket details wouldn't be cleaner. 

Regarding "I also see a lot of CDSE/CloudFerro references which we usually want to eliminate from the generic openeo-geopyspark-driver project". I guess the concern is mostly re-usability regardless of platform? I added 2 commits to deal with "unknown" providers such that you could still configure it like how it used to be and you don't need to know the provider. That way if it is a known provider users have to do less and it is easier to get started and for others they can configure it the old way or PR for support for another provider.

jdries (approved):
approved: this is quite urgent, I don't see major remarks.
We are progressively addressing other concerns as well, but not in this pr.

pvbouwel:
Closing in favor of #1192 since there the circular dependency is resolved

PR #1192, opened 2025-05-12
Workspace/allow cross region

pvbouwel:
Same as https://github.com/Open-EO/openeo-geopyspark-driver/pull/1190/commits

But version that keeps integrations.s3 as one


openeogeotrellis/integrations/s3/bucket_details.py
```diff
@@ -19,6 +18,8 @@ class BucketDetails:
 
     @classmethod
     def from_name(cls, bucket_name: str) -> BucketDetails:
+        from openeogeotrellis.workspace.object_storage_workspace import ObjectStorageWorkspace
```
pvbouwel, line 21: This is to avoid circular dependency and to avoid having to use s3_client as module name #1190

tests/conftest.py
```diff
@@ -450,21 +451,30 @@ def aws_credentials(monkeypatch):
...
+def mock_s3_resource(aws_credentials, mock_s3_client):
     if moto_server_address is None:
         with moto.mock_aws():
             yield boto3.resource("s3", region_name=TEST_AWS_REGION_NAME)
     else:
         yield boto3.resource("s3", region_name=TEST_AWS_REGION_NAME, endpoint_url=moto_server_address)
 
 @pytest.fixture(scope="function")
-def mock_s3_client(aws_credentials):
+def get_mock_s3_client(aws_credentials):
```
soxofaan, line 462: It's confusing to change the name of this fixture with a `get_` prefix:  
when using it in a test `get_mock_s3_client` will have the value of a client, not a getter of clients

tests/conftest.py
```diff
@@ -450,21 +451,30 @@ def aws_credentials(monkeypatch):
...
     else:
         yield boto3.client("s3", region_name=TEST_AWS_REGION_NAME, endpoint_url=moto_server_address)
 
+
+@pytest.fixture(scope="function")
+def mock_s3_client(get_mock_s3_client, monkeypatch):
+    def _get_client(*args, **kwargs):
+        return get_mock_s3_client
+
+    monkeypatch.setattr(S3ClientBuilder, "from_region", _get_client)
```
soxofaan, line 475: is this monkeypathing actually necessary? Doesn't the moto library support this out of the box?

soxofaan (commented):
some quick notest about fixture changes that threw me off a bit

tests/conftest.py
pvbouwel: For a lot of cases that is indeed true but we have test cases where the moto runs as the 'server' process but then a client mock should not follow ordinary moto mock logic but rather be a client configured to go to the endpoint of the moto server.

Issue #1015, opened 2025-01-22
Support new STAC "bands" construct


bossie:
`load_stac` only supports `eo:bands` but should support `bands` as well.

Case in point: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1014

m-mohr:
Important to note in this case are also the v2 versions of the raster, projection and eo extensions. Various renames happened there.

It might make sense to adopt an approach to migrate the STAC JSONs to the latest version first (e.g. through pystac or stac-migrate) so that you can work only on the latest version in the codebase. That has decreased the maintenance overhead in other tooling such as STAC Browser significantly in the past. stac-migrate is probably the most complete migration tool and has a CLI, but it runs in NodeJS. pystac comes next most likely.

soxofaan:
working on `load_stac` in geopyspark-driver feels a bit like an elephant in a china shop because of:
- https://github.com/Open-EO/openeo-geopyspark-driver/issues/1138

soxofaan:
related
- https://github.com/Open-EO/openeo-geopyspark-driver/issues/1109

JeroenVerstraelen:
@HansVRP mentioned that the data engineering team can help with testing the functionality for this ticket. To follow up.

HansVRP:
@manugv had a good suggestion for weed; we have a UDF/UDP that relies on loading their STACS: So having that one as an integration test would work well

soxofaan:
for clarity: PR to add support and some more related discussion is at
- https://github.com/Open-EO/openeo-geopyspark-driver/pull/1213

soxofaan:
setting this ticket as blocked, as there is a backward compatibility issue introduced by this feature request (LCFM use case related) and I'm not sure how to proceed

m-mohr:
How is this backward incompatible?

soxofaan:
> How is this backward incompatible?

it's not about (in)compatibilities on the level of the STAC API or STAC, 
but by adding support for STAC 1.1 bands in the geopyspark driver, the detected band names change e.g. :
- originally: "B01_20m" based on best effort fallback guess based on asset key
- now: "B01" based on STAC 1.1 band metadata (already present in the STAC resrouce)

so it's the behavior of the geopyspark driver that changes in an incompatible way 
e.g. uses have to change the band names they use in `filter_bands` or `array_element(label=...)`

jdries:
[...]

PR #1213, opened 2025-05-19
Add STAC 1.1 bands support to load_stac

soxofaan:
for #1015

self-review draft for now


openeogeotrellis/load_stac.py
```diff
@@ -368,8 +380,12 @@ def operator_value(criterion: Dict[str, object]) -> (str, object):
...
                     netcdf_with_time_dimension = contains_netcdf_with_time_dimension(collection)
 
-                band_names = [b["name"] for b in (catalog.summaries.lists if isinstance(catalog, pystac.Collection)
-                                                  else catalog.extra_fields.get("summaries", {})).get("eo:bands", [])]
+                if isinstance(catalog, pystac.Collection):
+                    band_names = _StacMetadataParser().bands_from_stac_collection(collection=collection).band_names()
+                elif isinstance(catalog, pystac.Catalog):
+                    band_names = _StacMetadataParser().bands_from_stac_catalog(catalog=catalog).band_names()
+                else:
+                    raise ValueError(catalog)
```
soxofaan, line 388: move this if-elif-chain into `_StacMetadataParser`

openeogeotrellis/load_stac.py
```diff
@@ -976,3 +992,88 @@ def single_filter(property, operator, value) -> dict:
...
+    def _band_from_common_bands_metadata(self, data: dict) -> openeo.metadata.Band:
+        """Construct band from metadata dict in STAC 1.1 + eo v2 style metadata"""
+        return openeo.metadata.Band(
+            name=data["name"],
+            common_name=data.get("eo:common_name"),
+            wavelength_um=data.get("eo:center_wavelength"),
+        )
+
+    def bands_from_stac_catalog(self, catalog: pystac.Catalog) -> _Bands:
+        # TODO: "eo:bands" vs "bands" priority based on STAC and EO extension version information
```
soxofaan, line 1034: add logging to have better insights what path is followed

tests/test_api_result.py
```diff
@@ -3927,12 +3979,14 @@ def item(path) -> dict:
...
 
-        urllib_and_request_mock.get("https://stac.test/collections/collection",
-                        data=get_test_data_file("stac/issue609-api-temporal-bound-exclusive/collection.json").read_text())
-        urllib_and_request_mock.get("https://stac.test",
-                        data=get_test_data_file("stac/issue609-api-temporal-bound-exclusive/catalog.json").read_text())
-        requests_mock.get("https://stac.test/search",
-                          json=feature_collection)
+        urllib_and_request_mock.get(
+            "https://stac.test/collections/collection",
+            data=get_test_data_file(f"{test_data_root}/collection.json").read_text(encoding="utf8"),
```
soxofaan: eliminate `get_test_data_file` here too

tests/test_api_result.py
```diff
@@ -3905,17 +3959,15 @@ def feature_collection(request, _) -> dict:
...
+                    preprocess=lambda s: s.replace(
                         "asset01.tiff",
                         f"file://{get_test_data_file('binary/load_stac/collection01/asset01.tif').absolute()}"
                     )
                 )
 
-            items = [item(path) for path in ["stac/issue609-api-temporal-bound-exclusive/item01.json",
-                                             "stac/issue609-api-temporal-bound-exclusive/item02.json",
-                                             ]]
+            items = [item(path) for path in [f"{test_data_root}/item01.json", f"{test_data_root}/item02.json"]]
```
soxofaan: simplify with `test_data` fixture

tests/test_load_stac.py
```diff
@@ -90,14 +94,22 @@ def feature_collection(request, _) -> dict:
...
     stac_api_root_url = "https://stac.test"
     stac_collection_url = f"{stac_api_root_url}/collections/collection"
 
-    stac_item = json.loads(
-        get_test_data_file("stac/issue609-api-temporal-bound-exclusive/item01.json")
-        .read_text()
-        .replace("asset01.tiff", f"file://{get_test_data_file('binary/load_stac/collection01/asset01.tif').absolute()}")
+    stac_item = test_data.load_json(
+        item_path,
+        preprocess=lambda s: s.replace(
```
soxofaan, line 110: use now built-in feature of TestDataLoader

tests/test_api_result.py
```diff
@@ -4031,17 +4106,14 @@ def item_json(path):
 
         res_path = tmp_path / "res.nc"
         res_path.write_bytes(res.data)
-
-        ds = xarray.load_dataset(res_path)
-
-        assert ds.dims["x"] == 12987
-        assert ds.dims["y"] == 767
+        da = _load_cube_from_netcdf(res_path)
```
soxofaan, line 4109: wider adoption of this new utility?

soxofaan:
about this PR

It's quite large in lines of code, but that is mainly because I parameterized a bunch of existing tests to have a legacy "eo:bands" variant and a STAC 1.1. common "bands" variant. This involved moving JSON files around and duplicating them with minimal changes, but that gives quite a substantial diff in git

To zoom in on the more relevant changes:

In `openeogeotrellis/load_stac.py`: I  extracted low level "eo:bands" detection from the main `load_stac` entry function and encapsulated that in a new helper `_StacMetadataParser`, where it is now easier to add support for the new STAC 1.1 "bands" construct.

`tests/test_load_stac.py` : there is one test `test_lcfm_improvements`  here were the added support for STAC 1.1 changes an outcome that might be important:
there is this original assert
https://github.com/Open-EO/openeo-geopyspark-driver/blob/6053a12518077473c26d698ae365a54ab359cdea/tests/test_load_stac.py#L204-L209
where `["B01_60m"]` is expected, 
but with the added support for "bands", this expectation is updated to `["B01"]`
because the related STAC doc explicitly defines this band name (and it explicitly declares to be a `"stac_version": "1.1.0"` doc):
https://github.com/Open-EO/openeo-geopyspark-driver/blob/6053a12518077473c26d698ae365a54ab359cdea/tests/data/stac/issue1043-api-proj-code/FeatureCollection.json#L282-L288

jdries:
This lcfm test change is quite important: turns out that the interpretation of band name is not that it maps to a unique asset.
What users want is to load a specific asset, by band name. It is fine to select the highest resolution one by default, but sometimes users indeed want to load B03_20m explicitly. Does that still work?

soxofaan:
> What users want is to load a specific asset, by band name. It is fine to select the highest resolution one by default, but sometimes users indeed want to load B03_20m explicitly. Does that still work?
> I believe we already have a system for 'aliases', so can we have one asset with name
> ['B01','B01_20m']
> and another one with ['B01_60m']

About that alias system: I first wrongly remembered this was a standardized feature for band harmonization  in *openEO* collections in a federation context. But I can only reconstruct that this is a VITO-specific trick to be more lenient about band name resolution (EP-3670, https://github.com/Open-EO/openeo-python-client/pull/188, https://github.com/Open-EO/openeo-geopyspark-driver/pull/69). So that worked because we have control over both the backend-side collection metadata and a power over the python client to include support for this non standardized trick.

To do the same in STAC will be a lot harder I guess.

soxofaan:
> What users want is to load a specific asset, by band name. It is fine to select the highest resolution one by default, but sometimes users indeed want to load B03_20m explicitly. Does that still work?

Just to be sure what you mean with "does that **still** work?":  
what now works is that users explicitly specify `B01_20m` or `B01_60m` and get the corresponding asset. But there is no feature that users can specify just `B01` and automatically get, e.g. `B01_20m`, right?

With this PR, and the current state of the STAC metadata of that use case, both `B01_20m` and `B01_60m` will collapse to the same band name `B01`, so that is a breaking change, indeed

soxofaan:
I have no idea if practically feasible for this use case, 
but isn't there a solution path were the "alias" aspect is encoded in a standardized way in the STAC metadata itself, 
e.g. roughly:

```json
  "assets": {
    "x": {  
      "href": "example.com/B01_20m",
      "bands": [{"name": "B01"}],
    },
    "y": {  
      "href": "example.com/B01_20m",
      "bands": [{"name": "B01_20m"}],
    },
    "z": {  
      "href": "example.com/B01_60m",
      "bands": [{"name": "B01_60m"}],
    },
```


note the duplicated href for band names B01 and B01_20m

this would allow the user pick a generic "B01" band if they don't care about resolution, or pick a specific one if they do care

openeogeotrellis/load_stac.py
```diff
@@ -458,7 +469,7 @@ def intersects_temporally(interval) -> bool:
                        .withNominalDate(itm.properties.get("datetime") or itm.properties["start_datetime"]))
 
             for asset_id, asset in band_assets.items():
-                asset_band_names = get_band_names(itm, asset) or [asset_id]
+                asset_band_names = get_band_names(item=itm, asset=asset) or [asset_id]
```
jdries, line 472: here we may need to check these band names inside an asset for overlap with other assets.
We now know that it is possible that two assets share a band name, so we need to resolve this before sending assets into geotrellis for datacube building. This seems like the best location.

Proposal: sort by gsd from smallest to larger, fallback to alphabetic order to make it deterministic?
Then keep a set of band names that have already been seen, and only add links for new names. Make sure to always include asset_id in list of band names.

soxofaan: I tried to implement this proposal in https://github.com/Open-EO/openeo-geopyspark-driver/pull/1213/commits/2d8422af2bc78001e02ca1467cd581b44ba1a3fc,
which reverts the change to the `test_lcfm_improvements` test

setup.py
```diff
@@ -60,7 +60,7 @@
     ],
     tests_require=tests_require,
     install_requires=[
-        "openeo>=0.33.0",
+        "openeo>=0.42.0.a2.dev",
```
soxofaan, line 63: 0.42 has been released in the meantime

soxofaan:
FYI: this  feature branch has a couple of commits:
-  to force the test suite to run a single test `test_load_stac_issue830_alternate_url` 
- disable error handling in web app layer to give full stack trace on failure

these commits should not be merged in master and are meant to be removed by rebase (or reverted)

Issue #1261, opened 2025-06-20
harden IPA lookup in batch job submit script


soxofaan:
The IPA lookup in the `submit_batch_job_spark3.sh` seems to be failing more recently

https://github.com/Open-EO/openeo-geopyspark-driver/blob/1c778e558dea393db5c39fefc675064e3cb43e6c/openeogeotrellis/deploy/submit_batch_job_spark3.sh#L107-L117

this bash construct is quite convoluted and hard to work with (e.g. how to retry when something goes wrong here?

better move this lookup to python, e.g. using the existing [freeipa integration](https://github.com/Open-EO/openeo-geopyspark-driver/blob/master/openeogeotrellis/integrations/freeipa.py)

related to https://github.com/Open-EO/openeo-geopyspark-driver/issues/627

jdries:
Note that there is already an 'if else' that falls back to using default principal rather than proxyuser. 
This can be extended to: if the request fails entirely, don't use proxyuser and still submit the job rather than failing.

soxofaan:
indeed that's what went for in PR https://github.com/Open-EO/openeo-geopyspark-driver/pull/1262

soxofaan:
ok tests pass now, and there is now a specific test for this feature.
If everything goes wrong, it reliably falls back to the original `run_as="--principal ${principal} --keytab ${keyTab}"`

PR #1262, opened 2025-06-20
Move IPA proxy user lookup to python

soxofaan:
for better traceability, error/fallback handling, etc

for
-  #1261


soxofaan:
ok tests pass now, and there is now a specific test for this feature.
If everything goes wrong, it reliably falls back to the original `run_as="--principal ${principal} --keytab ${keyTab}"`

PR #1272, opened 2025-06-27
resample_cube_spatial: estimate required partitions

jdries:
PR to apply same solution as for resample_spatial
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/478

TODO:
- apply the actual repartitioning
- test on snowcop downscaling example: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/477


jdries:
Assigning to @EmileSonneveld  as he also did the same for resample_spatial.
High priority for snowcop project as we want to run on long timeseries and large area there.

PR #1298, opened 2025-08-06
doc: job-options memory options

pvbouwel:
Clarify the documentation on memory related job-options. If the backend has a default of None then [a limit of 0 bytes is set](https://github.com/Open-EO/openeo-geopyspark-driver/commit/18e512d84c594591a90b2a357becfdc5f00266aa#diff-f606c26975b555ceec5f9fd97aa22cb9372fdd3d508db2eb8ae248f3b559f2eeL134) rather then no enforcement.

Also clarify that it is purely a limit and not a reservation and avoid mentioning UDFs for memory overhead.

### DIFF
--- a/openeogeotrellis/job_options.py
+++ b/openeogeotrellis/job_options.py
@@ -57,7 +57,7 @@ class JobOptions:
             "description": "Setting to specifically limit the memory used by python on a worker. "
             "Leaving this setting empty will enforce the default value of the backend. None will be zero bytes. "
             "Typical processes that use python-memory are UDF's, sar_backscatter or Sentinel 3 data loading. "
-            "This memory not a reservation so it can act as executor_memory_overhead but it is enforced as a limit."
+            "This memory is not a reservation so it can act as executor_memory_overhead but it is enforced as a limit."
             "Memory allocation problems with one of aforementioned processes likely warrant an increase of this value."
         })
     executor_cores: int = field(
--- a/openeogeotrellis/job_options.py
+++ b/openeogeotrellis/job_options.py
@@ -57,7 +57,7 @@ class JobOptions:
             "description": "Setting to specifically limit the memory used by python on a worker. "
             "Leaving this setting empty will enforce the default value of the backend. None will be zero bytes. "
             "Typical processes that use python-memory are UDF's, sar_backscatter or Sentinel 3 data loading. "
-            "This memory is not a reservation so it can act as executor_memory_overhead but it is enforced as a limit."
+            "This memory is not a reservation so it can act as executor_memory_overhead but it is enforced as a limit. "
             "Memory allocation problems with one of aforementioned processes likely warrant an increase of this value."
         })
     executor_cores: int = field(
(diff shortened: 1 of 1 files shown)


openeogeotrellis/job_options.py
```diff
@@ -48,15 +48,17 @@ class JobOptions:
...
-            "description": "Memory allocated to the workers in addition to the JVM, for example, to run UDFs. "
-            "The total available memory of an executor is equal to executor-memory + executor-memoryOverhead [+ python-memory].",
+            "description": "Memory allocated to the workers in addition to the JVM. "
+            "The total available memory of an executor is equal to executor-memory + executor-memoryOverhead + python-memory.",
         })
     python_memory: str = field(
         default=get_backend_config().default_python_memory,
         metadata={
             "description": "Setting to specifically limit the memory used by python on a worker. "
+            "Leaving this setting empty will enforce the default value of the backend. None will be zero bytes."
```
soxofaan, line 58: these lines are appended as is I think, so you have to put spaces at end (unless last sentence):
```suggestion
            "Leaving this setting empty will enforce the default value of the backend. None will be zero bytes. "
```

openeogeotrellis/job_options.py
```diff
@@ -48,15 +48,17 @@ class JobOptions:
...
             "Typical processes that use python-memory are UDF's, sar_backscatter or Sentinel 3 data loading. "
-            "Leaving this setting empty will allow Python to use almost all of the executor-memoryOverhead, but may lead to unclear error messages when the memory limit is reached."
+            "This memory not a reservation so it can act as executor_memory_overhead but it is enforced as a limit."
```
soxofaan, line 60: missing "is" here I think?

soxofaan: I'm also a bit confused by the "so it can act as executor_memory_overhead": 
does that bind to "reservation" or "not a reservation"?

Also: what is the (user-relevant) difference between being a reservation and a limit?

pvbouwel: A reservation means something is exclusively reserved and can only be used for that purpose so generally it acts as a lower bound in  terms of memory. A limit defines a limitation on a resource and thus in case of memory acts as an upper bound.

So "so it can act as executor_memory_overhead" binds to "not a reservation" because if you specify Python Memory to be 4GiB any other process can consume it hence it can be used as a "executor_memory_overhead"  as well so you should only use "executor_memory_overhead" if you want to have more memory and want to make sure it is not assignable to Python. Otherwise it is better to use Python memory as that can be for Python or anything else.

I am struggling however to have it in concise documentation and I thought the reservation and limit would help but it seems to cause confusion by itself .

soxofaan: Yeah, these option descriptions use terminology like "allocation", "reservation" and "limit", and apparently my Spark  memory management knowledge is a bit too rusty (like most users I guess) to understand the nuances.

jdries: Should we also say that we advice to work with python-memory rather than executor-memory-overhead?
The total memory is also a limited, but enforced by 'process killers' that simply kill the entire process, leading to hard-to-analyze debug messages.
The python-memory limit is enforced internally by python itself, which has 2 advantages:
- garbage collection and caching schemes can take the limit into account, and free memory when the limit is reached
- error messages clearly state when the error is caused by reaching the limit

Issue #1294, opened 2025-07-25
netCDF items no longer contain bbox and geometry in EJR results_metadata


JeroenVerstraelen:
With the recent changes for https://github.com/Open-EO/openeo-geopyspark-driver/issues/1111 it appears that we no longer store the bbox and geometry of netCDF assets in the `results_metadata` field of the EJR.

This means that the `/jobs/<job_id>/results/items/<item_id>` endpoint falls back to using the bbox of the entire job.

This breaks the WorldCereal extract_patches workflow since they create several patches per job and save these to netCDF files using `sample_by_feature`. Later, they upload these netCDF items to a STAC catalog which will now have the wrong geometry for each STAC item.

A good starting place to figure it where the regression started is this commit:
https://github.com/Open-EO/openeo-geopyspark-driver/pull/1141/commits/321bf7e3cd5a66313d8614419cc3725c557cb217

PR #1301, opened 2025-08-07
restore netCDF asset's bbox and geometry

bossie:
#1294


bossie:
Assigns `bbox` and `geometry` of items to their underlying assets as well (they should indeed correspond).

PR #1321, opened 2025-08-28
Better document colormap parameter

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/138

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -431,6 +431,17 @@ class GeoPySparkBackendImplementation(backend.OpenEoBackendImplementation):
         return get_backend_config().oidc_providers
 
     def file_formats(self) -> dict:
+        colormap_properties = {
+            "anyOf": [
+                {"type": "integer"},
+                {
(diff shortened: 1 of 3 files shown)


openeogeotrellis/backend.py
```diff
@@ -462,6 +473,7 @@ def file_formats(self) -> dict:
                 "GTiff": {
                     "title": "GeoTiff",
                     "description": "Cloud Optimized Geotiff is one of the most widely supported formats and thus a popular choice for further dissemination. This implementation stores all bands in one file, and creates one file per timestamp in your datacube.",
+                    "type": "object",
```
soxofaan, line 476: this addition has no meaning here I think

EmileSonneveld: True. I added it for the schema validation, but apparently, it is not needed https://github.com/Open-EO/openeo-geopyspark-driver/commit/3fede7d8c4a2ce5cd0537989b0ad80765f1c167f

Issue #1318, opened 2025-08-26
Set S3 attributes on upload to S3


JorisCod:
Set the following attributes on S3 upon export_workspace:

• md5: MD5 checksum die gebruik kan worden door CF om te verifiëren dat het product niet corrupt geraakt is tijdens het ingesten van de data
• mtime: wanneer data niet aangemaakt werd op het moment dat ze weggeschreven werd op S3, kan je in het mtime attribute de exacte timestamp aangeven (tov epoch, https://rclone.org/s3/#modification-times-and-hashes) wanneer de data aangemaakt werd.
--> Note: mtime is important, as we copy the product afterwards to give it the correct filename

(info from Dirk)

Addition from @JeroenVerstraelen, this is no longer blocked:
If I understand https://github.com/eu-cdse/openeo-cdse-infra/issues/460 correctly then CF no longer throws errors when you upload with the new boto3, but they both do this by ignoring the checksums so data integrity is not enforced.
Afaik this issue is not blocked anymore by the new boto3 version, and we can add the checksum and mtime attribute.

Further actions:
- [ ] Check how to set the checksum --> Similar to https://github.com/VITO-RS-Vegetation/elogs/blob/main/src/elogs/s3.py#L272
- [ ] Check on OTC

**Important:** seeing the limitations of s3fs, it seems best that this issue is only tackled after refactoring the code for uploading to S3 to use the S3 API instead of s3fs. (At least, this is what I understood after discussion with Jeroen and Peter).

bossie:
@JorisCod could you provide me with an ID of a workspace that you want to apply this to?

bossie:
@JorisCod this is available on staging; please test.

PR #1322, opened 2025-08-29
add "md5" and "mtime" metadata to assets exported as objects

bossie:
#1318


openeogeotrellis/workspace/helpers.py
```diff
@@ -43,7 +43,17 @@ def export_asset(asset: pystac.Asset, merge: PurePath, relative_asset_path: Pure
...
         if source_uri_parts.scheme in ["", "file"]:
-            s3_client().upload_file(str(source_path), asset_bucket, target_key)
+            s3_client().upload_file(
+                str(source_path),
+                asset_bucket,
+                target_key,
+                ExtraArgs={
+                    "Metadata": {
+                        "md5": md5_checksum(source_path),
+                        "mtime": str(time.time_ns()),
```
bossie, line 53: Get mtime from file instead.

PR #1364, opened 2025-09-26
calriassian: allow passing credentials for eodata

pvbouwel:


openeogeotrellis/integrations/calrissian.py
```diff
@@ -133,6 +196,7 @@ def __init__(
         self,
         *,
         namespace: str,
+        launch_config: CalrissianLaunchConfigBuilder,
```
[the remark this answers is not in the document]
soxofaan: indeed, in practice the `from_context()` factory is used, e.g.
https://github.com/Open-EO/openeo-geotrellis-kubernetes/blob/7f75ad3ee0a68be293875d29fc83e8d3b52e0797/openeo-geopyspark-k8s-custom-processes/src/openeo_geopyspark_k8s_custom_processes/custom_processes.py#L129

openeogeotrellis/integrations/calrissian.py
```diff
@@ -172,7 +237,36 @@ def __init__(
         )
 
     @staticmethod
-    def from_context() -> CalrissianJobLauncher:
+    def get_eodata_access_env_vars(user_id: str, correlation_id: str) -> dict:
```
soxofaan, line 240: general thought: what does "eodata" mean here actually in this context? 
It's used here in a method name, which might be too much coupling with some implementation detail on CDSE?

pvbouwel: Earth observation data.  What that exactly means depends on the platform. This configuration assumes anyway that the platform deploys an S3Proxy because it uses identity tokens that only makes sense against such proxy. The proxy is configured to forward requests to an S3 API but which one that is a configuration of that proxy. 

As such it can be any S3 resource but one where you expect the access to be provided by the platform. 

Would it be clearer if it is `get_s3proxy_access_env_vars` ?

PR #1372, opened 2025-09-30
test fix with original job

bossie:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/446


bossie:
Tests, ran against the jars produced by https://github.com/Open-EO/openeo-geotrellis-extensions/pull/544, still succeed* so closing.

*exluding those that test/rely on an x.y.z version schemes of said jars.

Issue #1360, opened 2025-09-22
save_result: allow results larger than 5GB if driver-memory is available


jdries:
This line:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/1e0ad3ad2082c9ae2244d18015808adb33efd2d0/openeogeotrellis/deploy/sparkapplication_template.yaml.j2#L56

puts a hard-coded limit of 5GB on certain results, for instance, geotiff without time dimension.

When the user configures driver-memory, we could increase this limit accordingly, allowing to export larger results.

This issue blocks writing larger results, and is very easy to fix. Setting priority to medium for now, but can become high.

PR #1398, opened 2025-10-28
feat: allow results larger than 5GB if driver memory is available

dsamaey:
# https://github.com/Open-EO/openeo-geopyspark-driver/issues/1360

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -71,6 +71,7 @@ from pandas import Timedelta
 from py4j.java_gateway import JVMView
 from py4j.protocol import Py4JJavaError
 from pyspark import SparkContext
+from pyspark.util import _parse_memory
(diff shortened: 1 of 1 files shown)


openeogeotrellis/backend.py
```diff
@@ -2024,6 +2024,11 @@ def as_boolean_arg(job_option_key: str, default_value: str) -> str:
 
             batch_job_cfg_secret_name = k8s_get_batch_job_cfg_secret_name(spark_app_id)
 
+            max_result_size = '5g'
+            from pyspark.util import _parse_memory
+            if _parse_memory(max_result_size) < _parse_memory(options.executor_memory):
+                max_result_size = options.executor_memory
```
jdries: I believe you'll want to check against driver_memory, the max_result_size is what gets collected into the driver, so the combination of all the data in the partitions stored on executors.

PR #1411, opened 2025-11-14
ItemCollection.deduplicate

### LINKED ISSUE #1361: load_stac: item deduplication
In production catalogs, the same item can occur with multiple versions.
The scala based catalog client has a 'deduplication' feature to select the most recently updated item.

In CGLS collections we have something similar with reprocessed products receveing a property with value RT0 -RT1 -... - RT6. Here the highest 'RT' item needs to be selected.

Scala implementation:
https://github.com/Open-EO/openeo-opensearch-client/blob/485ee07345822ce3a29899e4ce7fd890e0003bf8/src/main/scala/org/openeo/opensearch/OpenSearchResponses.scala#L246

soxofaan:
resolves #1361

### DIFF
--- a/openeogeotrellis/load_stac.py
+++ b/openeogeotrellis/load_stac.py
@@ -95,13 +95,17 @@ def load_stac(
     batch_jobs: Optional[openeo_driver.backend.BatchJobs] = None,
     override_band_names: Optional[List[str]] = None,
     stac_io: Optional[pystac.stac_io.StacIO] = None,
+    feature_flags: Optional[Dict[str, Any]] = None,
 ) -> GeopysparkDataCube:
     if override_band_names is None:
         override_band_names = []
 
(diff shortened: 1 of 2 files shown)


soxofaan:
- [x] add changelog entry
- [x] bump version number

openeogeotrellis/load_stac.py
```diff
@@ -211,6 +214,13 @@ def load_stac(
 
                 item_collection = ItemCollection.from_stac_catalog(catalog, spatiotemporal_extent=spatiotemporal_extent)
 
+        # Deduplicate items
+        # TODO: smarter and more fine-grained deduplication behavior?
+        #       - enable by default or only do it on STAC API usage?
+        #       - allow custom deduplicators (e.g. based on layer catalog info about openeo collections)
+        if feature_flags.get("deduplicate-items", False):
```
soxofaan: rename to "deduplicate_items" for better consistency with other feature flags and parameters

PR #1434, opened 2025-11-26
add docstring to load_stac function

JeroenVerstraelen:
Added a docstring to the load_stac function to clarify its usage.

The docstring focuses on aspects that weren’t immediately clear when reading the code for the first time, with the goal of helping users understand the function and its parameters without needing to dive into the implementation. I hope it can also help during refactoring to keep a bird's eye view of the total inputs, outputs, and goals of the function.

### DIFF
--- a/openeogeotrellis/load_stac.py
+++ b/openeogeotrellis/load_stac.py
@@ -97,6 +97,81 @@ def load_stac(
     stac_io: Optional[pystac.stac_io.StacIO] = None,
     feature_flags: Optional[Dict[str, Any]] = None,
 ) -> GeopysparkDataCube:
+    """
+    Load a STAC resource (Collection, Item, Catalog, or batch job results) and return
+    it as a GeopysparkDataCube. Based on filters defined in load_params and layer_properties.
+
+    :param url: STAC URL to load from. Can point to:
+        - **STAC Collection**:
+          Items are queried from the STAC API that intersect the requested
+          spatiotemporal extent.
+
+        - **STAC Item**:
+          A single item is used if it intersects the requested spatiotemporal
+          extent.
+
+        - **STAC Catalog**:
+          The catalog and all of its subcatalogs are recursively searched for
+          items that intersect the requested spatiotemporal extent.
+          Multiple collections may contribute items to the resulting cube.
+
+        - **Dependency Job**:
+          URLs of the form `*/jobs/{job_id}/results`. If `user` and
+          `batch_jobs` are provided and the user owns the job, the job results
+          catalog is fetched. While fetching, it will wait for the job to
+          complete if necessary (up to a configured maximum).
+          Only items from the job results catalog intersecting the requested
+          extent are included.
+
+    :param load_params: Parameters and optimization hints for
+        load_collection/load_stac, derived from dry-run source constraints.
+        It contains the spatiotemporal extent provided by the user.
+        It also contains the global_extent, target_crs, target_resolution,
+        bands, properties, and aggregate_spatial_geometries parameters.
+        Note that the global_extent is divided into spatial keys and it is
+        important that it is aligned properly with the target resolution.
+
+    :param env: Evaluation environment containing user information, batch job
+        options, process-graph parameter values used for property filtering, etc.
+
+    :param layer_properties: Properties from the layer catalog's
+        `_vito.properties` field used to filter STAC items based on metadata
+        fields. These are merged with `load_params.properties` provided by the user.
+        Example filter (keeping only STAC items with `properties.season == "s1"`):
+
+            "properties": {
+                "season": {
+                    "process_graph": {
+                        "eq1": {
+                            "process_id": "eq",
+                            "arguments": {
+                                "x": {"from_parameter": "value"},
+                                "y": "s1"
+                            },
+                            "result": true
+                        }
+                    }
+                }
+            }
+
+    :param batch_jobs: Optional job registry used to resolve dependency-job
+        result URLs.
+
+    :param override_band_names: Optional list of band names to enforce in the
+        resulting data cube metadata. Currently only provided by the layer
+        catalog to enforce specific band names.
+
+    :param stac_io: Optional custom STAC I/O handler (e.g., to enable
+        authentication or caching).
+        TODO: Currently never provided in any load_stac call.
+
+    :param feature_flags: Optional flags to modify load_stac behavior.
+        TODO: Currently never provided in any load_stac call.
+
+    :return: A `GeopysparkDataCube` loaded from the provided STAC URL containing
+        only the items intersecting the requested spatiotemporal extent and the
+        requested property values.
+    """
     if override_band_names is None:
         override_band_names = []
 


openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
     stac_io: Optional[pystac.stac_io.StacIO] = None,
     feature_flags: Optional[Dict[str, Any]] = None,
 ) -> GeopysparkDataCube:
+    """
+    Load a STAC resource (Collection, Item, Catalog, or batch job results) and return
```
soxofaan:
```suggestion
    Load a STAC resource (STAC Collection/Catalog, STAC Item, STAC API, or openEO batch job results) and return
```

openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
...
+    it as a GeopysparkDataCube. Based on filters defined in load_params and layer_properties.
+
+    :param url: STAC URL to load from. Can point to:
+        - **STAC Collection**:
```
soxofaan:
```suggestion
        - **STAC API**:
```

openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
...
+          The catalog and all of its subcatalogs are recursively searched for
+          items that intersect the requested spatiotemporal extent.
+          Multiple collections may contribute items to the resulting cube.
+
+        - **Dependency Job**:
+          URLs of the form `*/jobs/{job_id}/results`. If `user` and
+          `batch_jobs` are provided and the user owns the job, the job results
+          catalog is fetched. While fetching, it will wait for the job to
+          complete if necessary (up to a configured maximum).
+          Only items from the job results catalog intersecting the requested
```
soxofaan:
```suggestion
          Only assets from the job results intersecting the requested
```

openeogeotrellis/load_stac.py
JeroenVerstraelen: If I understand the STAC spec correctly, there are [three main components/resources](https://stacspec.org/en):
* Catalog (dynamic or static)
* Collection
* Item

I understand that a Collection is a Catalog with additional fields but I think we can benefit from always seeing a Collection and Catalog as two different objects. In our current load_stac code we have a clear structure:
```
if isinstance(stac_object, pystac.Item)
elif isinstance(stac_object, pystac.Collection)
else:
    assert isinstance(stac_object, pystac.Catalog) 
```

The STAC Specification consists of 4 semi-independent specifications:
* STAC Catalog
* STAC Collection
* STAC Item
* STAC API: RESTful API defined in https://api.stacspec.org/

Our `load_stac` function only supports the first three specifications. Our code nor Pystac has any checks for the `url` parameter. So it can be any endpoint, as long as it returns json and Pystac can parse it as a STAC item, collection or catalog.

In that sense we can just reference the three different types of payloads we support:
`Load a STAC resource (Collection, Item, Catalog) from any JSON endpoint and convert the STAC items into a GeopysparkDataCube. Based on item filters defined in load_params and layer_properties.`

We can keep the `Own openEO Batch Job` reference for the documentation of the `url` parameter. It is the only type of URL where we do an explicit form check (`*/jobs/{job_id}/results`). It should be transparent to the user of the function that we don't actually extract the payload from this URL, but we take it directly from the jobregistry.

JeroenVerstraelen: So my suggestion here is:
```
Load a STAC JSON object (Collection, Item, Catalog) from any JSON endpoint and convert the STAC items into a GeopysparkDataCube. Based on item filters defined in load_params and layer_properties.
```

openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
...
+        - **STAC Item**:
+          A single item is used if it intersects the requested spatiotemporal
+          extent.
+
+        - **STAC Catalog**:
+          The catalog and all of its subcatalogs are recursively searched for
+          items that intersect the requested spatiotemporal extent.
+          Multiple collections may contribute items to the resulting cube.
+
+        - **Dependency Job**:
```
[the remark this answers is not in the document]
JeroenVerstraelen: That's better indeed, I think we should also rename "dependency job" in the code because when I first saw just the name without the implementation it confused me at first.

openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
...
+                                "x": {"from_parameter": "value"},
+                                "y": "s1"
+                            },
+                            "result": true
+                        }
+                    }
+                }
+            }
+
+    :param batch_jobs: Optional job registry used to resolve dependency-job
```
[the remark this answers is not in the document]
JeroenVerstraelen: But in the code `batch_jobs` is only used to get information from the job registry. 
```
batch_jobs.get_job_info(job_id=job_id, user_id=user_id)
batch_jobs.get_result_assets(job_id=job.id, user_id=user.user_id)
```

We might have to pass something smaller here. Because we're passing an object with the following interface, while we only need some basic metadata from the job. As far as a user of `load_stac` can tell, someone could call `delete_job` or `delete_batch_process_results` when he passes this object.
```
__init__
set_default_sentinel_hub_credentials
set_terrascope_access_token_getter
create_job
get_job_info
poll_job_dependencies
get_user_jobs
get_job_output_dir
get_job_work_dir
start_job
_start_job
_determine_container_image_from_process_graph
_schedule_and_get_dependencies
get_result_metadata
get_result_assets
_results_metadata_to_assets
get_results_metadata_path
load_results_metadata
_load_results_metadata_from_file
_get_providers
get_log_entries
cancel_job
delete_job
_delete_job
delete_batch_process_results
delete_batch_process_dependency_sources
```

openeogeotrellis/load_stac.py
```diff
@@ -97,6 +97,81 @@ def load_stac(
...
+
+        - **STAC Catalog**:
+          The catalog and all of its subcatalogs are recursively searched for
+          items that intersect the requested spatiotemporal extent.
+          Multiple collections may contribute items to the resulting cube.
+
+        - **Dependency Job**:
+          URLs of the form `*/jobs/{job_id}/results`. If `user` and
+          `batch_jobs` are provided and the user owns the job, the job results
+          catalog is fetched. While fetching, it will wait for the job to
```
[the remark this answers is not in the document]
soxofaan: > it will literally fetch the job results

Maybe you don't mean it like that, but that sounds like it actually downloads all the assets at that point, which is not the case.
It just gets/polls the batch job result _metadata_ document (`GET /jobs/{job_id}`) to get the job status and actively waits until that status reaches finished/error. No "results" are fetched before that (there aren't any yet anyway).


Also: at the moment, the logic doesn't even assume the metadata document is a STAC _collection_ (to be compatible with openeo<1.0 I think). The only assumption is that there are STAC assets linked from the document. 
There is also no assumption about STAC items or STAC item links

openeogeotrellis/load_stac.py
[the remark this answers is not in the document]
soxofaan: yes, I'd also like to rework the "own job" flow a bit, 
but as mentioned, this is the least relevant/urgent part of the load_stac refactor, so I would just stick to documenting the current state

PR #1328, opened 2025-09-02
[APEx] Binning implementation for Sentinel-3 collection loading for EFAST

Hneuschmidt:
This PR introduces a new resampling method, "binning" to the `load_collection` process for Sentinel-3 collections.
The main additions are the following:

1. The binning algorithm itself (mean aggregation) in `binning.py`
2. Changes to sentinel3.py, switching between the existing `"nearest_neighbor"` reprojection. The reprojection type (`"binning"` or `"nearest_neighbor"`) is read from the `featureflags` attribute which can be passed to `load_collection` using [Advanced Argument tweaking](https://open-eo.github.io/openeo-python-client/processes.html#advanced-argument-tweaking). 

Please create a new branch for this pull request.
 
## Explanation of changes

In `sentinel3.py`: the parameters are read from `feature_flags` in `pyramids` and passed down to `read_product` and `create_s3_toa` where we dispatch either `do_reproject` or the new `do_binning` depending on the value of `reprojection_type`.

It should be noted that while the documentation uses the example of a `"feature_flags"` parameter (this is also the name of the parameter to `pyramids`, the code expects `"featureflags"` without the underscore in [ProcessGraphDeserializer.py](https://github.com/Open-EO/openeo-python-driver/blob/0d4f99c50671ceb0ad26ec8f1c431f9d9696cb55/openeo_driver/ProcessGraphDeserializer.py#L783). Therefore, a user would enable binning for load collection in a line similar to this:

```Python
conn =  openeo.connect("https://openeo.dataspace.copernicus.eu/").authenticate_oidc()
cube = conn.load_collection("SENTINEL_SYN_L2_SYN", spatial_extent, temporal_extent
cube.result_node().update_arguments(featureflags=dict(reprojection_type="binning", super_sampling=3))
```

The binning algorithm is provided standalone so that it could be integrated into other processes (e.g. resample_spatial) later.

## Binning

Unlike nearest neighbor resampling, which assigns each output pixel the input pixel from which to take the value, binning operates in the opposite direction. It determines the output pixel in which the pixel center of an input pixel falls. If an output pixel is assigned multiple input pixels, these must be aggregated. In this PR, mean aggregation is the only aggregation method that is implemented. The binning implementation here is based on (and much less general than) the Binning Operator from SNAP [1], which in turn is based on the NASA SeaWiFS binning algorithm, as explained in [1].

[...]

PR #1458, opened 2025-12-11
Bc binning mask extension

Hneuschmidt:
This PR extends the "binning" type resampling for Sentinel-3 collections introduced in #1328 with the ability to mask out invalid pixels.

## Why is this needed?

The binning implementation only supports mean binning, where all contributions from input pixels are averaged in the output pixel. This works well for radiance values but not for flags (e.g the CLOUD_flags band of `SY_2_SYN___` products, as the average value of the quality flags is meaningless.

Currently, the OpenEO EFAST implementation, for which the binning feature was developed uses nearest neighbor (NN) resampling for the CLOUD_flags band. This achieves reasonable results for the flag band itself (nearest neighbor is adequate to resample flag values), but creates inaccuracies in the final result. The binned data bands include values in neighboring pixels that should have been masked out before the procedure and the NN based mask applied after binning does not correspond exactly to the input pixels, due to the resampling.

The accurate way to mask invalid (cloudy) values for binning is to apply the mask before the binning is done, so that these values cannot contribute to any of the output pixels. This PR adds a feature (hidden behind the feature_flags mechanism) to support this use case.

## How does the implementation work?

The previous PR added two parameters that are read from `feature_flags` passed to the `load_collection` process:

1. `reprojection_type` to switch between nearest neighbor and binning resampling
2. `super_sampling` to control super sampling for binning.

This PR adds two more parameters to configure binning:

1. `flag_band` to specify a band as a basis for the mask
2. `flag_bitmask` to specify a bitmask that is compared with `flag_band` for each pixel, before binning.

Both of the parameters are optional and have sensible defaults:

1. `flag_band` is None by default, meaning no masking is done
2. `flag_bitmask` is 0xff by default, masking out any pixel where flag_band is not zero

In the binning procedure, the flag band is treated specially (if the parameter `flag_band` is set): Pixels of any of the other bands specified for `load_collection` are set to `NaN` after loading the data from the SAFE archive. They are then assigned a weight of zero (ignored) in the averaging step of the binning procedure.

### DIFF
--- a/openeogeotrellis/collections/sentinel3.py
+++ b/openeogeotrellis/collections/sentinel3.py
@@ -452,7 +452,7 @@ def do_binning(
         angle_data_mask,
         digital_numbers=True,
         super_sampling: int=DEFAULT_SUPER_SAMPLING,
-        flag_band: str | None=None,
+        flag_band: str=None,
         flag_bitmask: np.uint16=DEFAULT_FLAG_BITMASK,
     ):
     """


jdries (approved):

jdries:
Can be merged as soon as CI succeeds (awaiting results).

Issue #1302, opened 2025-08-07
load_stac: support loading CDSE sentinel2-l2a


JeroenVerstraelen:
sentinel2-l2a is a very good test case for advanced load_stac. We should test this after the refactor, and check the following:

- loading of only 60m assets should result in a 60m cube
- loading of mixed resolutions should also work, and use highest resolution
- offsets to digital numbers should be applied correctly

jdries:
Indeed, users really should not try to load from this (particularly complex) STAC collection, but rather use the official openEO integrated collection.
It is certainly the idea that we fully switch also ourselves to using this one, but that is part of the whole load_stac refactor issue.

jdries:
Raised priority: this is not just a test case but a very important issue now that opensearch api will be shut down soon.

soxofaan:
> offsets to digital numbers should be applied correctly

The problem is what "correctly" means.

- User-interface-wise it's probably most correct to automatically apply offset *and* scale to give users physical values instead of digital numbers
- But because of performance considerations we currently only apply the offset, but not the scale in `SENTINEL2_L2A` and related collections


Related: https://github.com/Open-EO/openeo-geopyspark-driver/commit/835eb5205ce198a3f7d502d1668ff98f82a3c6b0  (to only do the offset-only transform on reflectance bands, but not WVP, AOT, ...)

PR #1491, opened 2026-01-08
Support granule_metadata asset in SENTINEL2_L2A for azimuth/zenith angle bands

soxofaan:
refs eu-cdse/openeo-cdse-infra#804, #1302, eu-cdse/openeo-cdse-infra#802, eu-cdse/openeo-cdse-infra#799


jdries:
Looks good enough to merge.
Remarks that I would not address before merging:
- you can actually set up a test with a real metadata file, maybe even some multi-resolution jp2's, that would give you faster feedback on the complex sentinel-2 collection
- the custom behavior could be encapsulated, so that it's only available for specific collections and not in the main code path. The item/asset augmentation feature for instance could add these virtual bands.

jdries (approved):

soxofaan:
thanks, merged in 6deb11193c048c8c176955072b05ab2d9142d723

> custom behavior could be encapsulated, so that it's only available for specific collections and not in the main code path.

note that the feature requires an explicit feature flag from layercatalog.json to be set, so in that sense it should not affect the default main code path

PR #1498, opened 2026-01-09
calrissian: specify container resources for all utility pods

pvbouwel:
Make sure calrissian utility pods will have resource specifications.

For pods that stage inputs the memory footprint is small and the pods are very short-lived so it is OK to hard-code a value deemed safe for any pratical case. (Since input is base64 passed it has anyway length constraints).

For the pod running calrissian itself the config is not as clear cut. We go for 256MB which should be a comfortable default but we take if from config such that it can be experimented with and backend can tune it for their supported workflow use cases.


EmileSonneveld:
@codex review

PR #1500, opened 2026-01-13
AdaptingPropertyFilter

soxofaan:
for the property filter mapping feature discussed in 
- https://github.com/eu-cdse/openeo-cdse-infra/issues/804


soxofaan:
@JeroenVerstraelen 
review-wise it's enough to just review commit f2afdadfb10b5b19223d6a7352a0e99758b20a18
the other commit is just a supporting/preparing refactor commit without functional change

soxofaan:
Example of what would have to be added as `load_stac_feature_flags` in layercatalog.json for `SENTINEL2_L1C_STAC` and `SENTINEL2_L2A_STAC `:

```json
"property_filter_adaptations": {
            "productType": "drop",
            "tileId": {
              "rename": "grid:code",
              "value_mapping": "add-MGRS-prefix"
            }
          }
```

JeroenVerstraelen:
@soxofaan I checked the implementation and it LGTM, the mapping is exactly what we discussed and I see no issues with it. We have room to expand it if other collections ever need prefix mapping etc.

PR #1502, opened 2026-01-15
810 cdse migrate sentinel1 grd to stac

JeroenVerstraelen:


openeogeotrellis/collections/s1backscatter_orfeo.py
```diff
@@ -125,6 +126,132 @@ class S1BackscatterOrfeo:
...
+                jvm.org.openeo.opensearch.OpenSearchResponses.featureBuilder()
+                .withNominalDate(itm.properties.get("datetime") or itm.properties["start_datetime"])
+                .withGeometryFromWkt(str(shapely.geometry.shape(itm.geometry)))
+            )
+            if not itm.bbox:
+                raise OpenEOApiException(f"Item {itm.id} has no bbox")
+            latlon_bbox = BoundingBox.from_wsen_tuple(itm.bbox, 4326)
+            builder = builder.withBBox(*map(float, latlon_bbox.as_wsen_tuple()))
+
+            # Extract product ID from asset href
```
jdries: why not use item.id ? Do we know where product id is being used?

JeroenVerstraelen:
Added deprecation warning. For example:
`WARNING Deprecated property 'productType'='IW_GRDH_1S-COG' was automatically converted to new STAC property 'product:type'='IW_GRDH_1S_B'. Please update your process graph to use the new property and value in the future.`

PR #1505, opened 2026-01-15
load_collection through load_stac: support band name aliases/variants

soxofaan:
refs: eu-cdse/openeo-cdse-infra#804, eu-cdse/openeo-cdse-infra#803


soxofaan:
original PR didn't work
had to do fixup dd913da4354452b508fb45adc51ddcdf77aeb4e7

PR #1509, opened 2026-01-16
Feature/billable metrics

pvbouwel:
Allow emitting billable metrics. These are metrics that are scraped and based on the pod requests because those are reservations made for the workload and which cannot be used by other workloads.

To enable logging the environment variable `LOG_BILLABLE_METRICS` needs to be toggled. It is also possible to account for missed time due to scraping by setting `CORRECTION_SCRAPE_TIME_SECS` to the scrape time.

These code changes will not change the metrics communicated with the ETL api. At this stage the new metrics will only be logged.

### DIFF
--- a/openeogeotrellis/integrations/prometheus.py
+++ b/openeogeotrellis/integrations/prometheus.py
@@ -85,29 +85,29 @@ class SingleTimeSeriesFloatStats:
         min_val: Optional[float] = None
         max_val: Optional[float] = None
         weighted_sum: Optional[float] = 0.0
-        prev_point: Optional[T_PromDatapoint] = None
+        prev_typed_point: Optional[T_PromDatapoint] = None
 
         for point in datapoints:
             point_ts = point[0]
-            point_value = float(point[1])
+            point_float_value = float(point[1])
 
             if start is None:
                 start = point_ts
 
             if min_val is None:
-                min_val = point_value
+                min_val = point_float_value
             else:
-                min_val = min(min_val, point_value)
+                min_val = min(min_val, point_float_value)
 
             if max_val is None:
-                max_val = point_value
+                max_val = point_float_value
             else:
-                max_val = max(max_val, point_value)
+                max_val = max(max_val, point_float_value)
 
-            if prev_point is not None:
-                weighted_sum += float(prev_point[1]) * (point_ts - prev_point[0])
+            if prev_typed_point is not None:
+                weighted_sum += float(prev_typed_point[1]) * (point_ts - prev_typed_point[0])
 
-            prev_point = tuple(point_ts, point_value)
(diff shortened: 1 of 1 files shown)


openeogeotrellis/integrations/prometheus.py
```diff
@@ -1,6 +1,154 @@
...
+        elif self.start == self.end:
+            # Single sample so no duration available
+            assert self.min == self.max
+            return self.min * scrape_time_seconds
+        else:
+            duration_seconds = self.end - self.start
+            return self.weighted_sum * ((duration_seconds + scrape_time_seconds) / duration_seconds)
+
+
+def time_series_to_single_float_value(data: T_PromQueryRespTimeSeries, *, compactable: bool = False) -> float:
```
JeroenVerstraelen: This name might be too generic. Could something like `sum_weighted_timeseries` or `sum_timeseries` be better? It is more in line with other function names such as `compact_time_series`.

openeogeotrellis/integrations/prometheus.py
```diff
@@ -1,6 +1,154 @@
...
+        _log.debug("Compacted data", extra={"data": data})
+    summed_value = 0.0
+    for timeserie in data:
+        metric = timeserie["metric"]
+        stats = TimeSerieFloatStats.from_timeserie(timeserie["values"])
+        _log.debug(
+            f"Summed value for {metric.get('resource', 'unknown')}",
+            extra={"pod": metric.get("pod", "unknown"), "unit": metric.get("unit", "unknown"), "stats": stats},
+        )
+        summed_value += stats.corrected_weighted_sum
```
JeroenVerstraelen: stats can be None here.

openeogeotrellis/integrations/prometheus.py
```diff
@@ -64,12 +236,69 @@ def get_max_executor_memory_usage(self, application_id: str, at: str = None) ->
         query = f'max(max_over_time(container_memory_working_set_bytes{{pod=~"{application_id}-.+exec.+",image=""}}[5d]))/(1024*1024*1024) '
         return self._query_for_float(query, at)
 
+    def get_summed_float_values(
+        self, query: str, job_id: str, start: float, end: float, *, compactable: False
```
JeroenVerstraelen: `: False`, I think you meant `= False` here.

EmileSonneveld:
@codex review

openeogeotrellis/integrations/prometheus.py
pvbouwel: Will go for `sum_timeseries_weighted`

EmileSonneveld (approved):
LGTM. Especially because everything is behind the `LOG_BILLABLE_METRICS` flag

Issue #1523, opened 2026-01-28
Improve requirements for `load_stac`able resources


soxofaan:
We already have an initial listing of requirements for STAC resources supported by `load_stac`

https://github.com/Open-EO/openeo-geopyspark-driver/blob/27105ee09584b892a41ac11ab2edb5b9637227d4/docs/load_stac.qmd#L41-L46

but this description is not in line with current implementation

needs fine-tuning

PR #1524, opened 2026-01-28
docs: fine-tuning of load_stac requirements

soxofaan:
for 
-  #1523


soxofaan:
cc @przell

soxofaan:
Note that the parsing logic this description is based on is currently defined in the openeo-python-client:
https://github.com/Open-EO/openeo-python-client/blob/d97d8fea79034446198a613a8b821101a1aa1234/openeo/metadata.py#L855-L941

it's currently (intentionally) a private helper, experimental and not for public use yet, 
but with appropriate polishing this could turn in:
- a tool for user-side validation/simulation of how back-end side `load_stac` might parse the STAC resource
- reusable utility to use across multiple backend implementations for more conistency

PR #1561, opened 2026-02-26
Test request too much

EmileSonneveld:
@codex review
https://github.com/cloudinsar/s1-workflows/issues/87


EmileSonneveld:
Extra test that was done, that is not clean enough to commit:
```python
path = "/.../cwltool/tests/"
for p in Path(path).glob("**/*.cwl"):
    print(p)
    try:
        cwl = CwLSource.from_path(p)
    except Exception as e:
        # if the CWL is no valid yaml, we ignore it
        continue
    max_memory = cwl.estimate_max_memory_usage()
    print(f"{max_memory=}")

```

Issue #1565, opened 2026-02-27
load_stac: post_dry_run handling takes 15 minutes per source_constraint (450 STAC items)


JeroenVerstraelen:
In these logs you can see that it appears to take 15 minutes for the post_dry_run to handle one source_constraint. This is a major blocker for EUGW production workflows.

[Kibana logs](https://kibana-infra.vgt.vito.be/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2026-02-27T13:16:12.487Z',to:'2026-02-27T13:57:20.835Z'))&_a=(columns:!(message,levelname,filename,kubernetes.container.name),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'592a42f0-e665-11ec-8cc4-3747d5233c59',key:filename,negate:!t,params:(query:jobregistry.py),type:phrase),query:(match_phrase:(filename:jobregistry.py))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'592a42f0-e665-11ec-8cc4-3747d5233c59',key:filename,negate:!t,params:(query:job_tracker_v2.py),type:phrase),query:(match_phrase:(filename:job_tracker_v2.py))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'592a42f0-e665-11ec-8cc4-3747d5233c59',key:filename,negate:!t,params:(query:util.py),type:phrase),query:(match_phrase:(filename:util.py)))),index:'592a42f0-e665-11ec-8cc4-3747d5233c59',interval:auto,query:(language:kuery,query:'job_id%20:%20%22j-2602271303304bf98c3fd29ed5664c92%22%20'),sort:!(!('@timestamp',desc))))

```
Feb 27, 2026 @ 14:20:13.363	load_stac of arbitrary URL https://stac.dataspace.copernicus.eu/v1/collections/sentinel-2-l2a	INFO	load_stac.py	

Feb 27, 2026 @ 14:20:13.554	No cube:dimensions metadata	WARNING	metadata.py	

[...]

PR #1567, opened 2026-02-28
add caching to _extract_spatial_extent_from_constraint_load_stac

soxofaan:
for #1565


soxofaan:
probably not going to merge this. There is by now already caching in a couple of other places

however, some parts of this PR are interesting to keep (e.g. added hashability of some objects), so I'm going to cherry pick those separately

PR #1661, opened 2026-04-28
ETL API: report added value for sync requests

### LINKED ISSUE #1436: ETL API: report added value for sync requests
Requested by @JanssenBrm: like batch jobs, sync requests should report added value in addition to resource usage.

The information that we currently lack in at [this point in the code](https://github.com/Open-EO/openeo-geopyspark-driver/blob/03243eb2b0001693e69efd96dff68d2b572159c3/openeogeotrellis/backend.py#L1279-L1340) is:
- area
- processes called

bossie:
#1436


openeogeotrellis/backend.py
```diff
@@ -1302,61 +1305,73 @@ def request_costs(
         job_options: Union[dict, None] = None,
         request_id: str,
         success: bool,
+        process_graph: Union[ProcessGraphFlatDict, None] = None,
+        tracer: Union[DryRunDataTracer, None] = None,
```
soxofaan, line 1309: `request_costs` is called from openeo_driver, so this depends on a change there, right?

soxofaan: ok apparently this
- https://github.com/Open-EO/openeo-python-driver/pull/495

PR #1676, opened 2026-05-21
fix idempotency key for added value request

### LINKED ISSUE #1436: ETL API: report added value for sync requests

bossie:
#1436

### DIFF
--- a/openeogeotrellis/integrations/etl_api.py
+++ b/openeogeotrellis/integrations/etl_api.py
@@ -226,7 +226,7 @@ class EtlApi:
             "orchestrator": ORCHESTRATOR,
             "jobStart": started_ms,
             "jobFinish": finished_ms,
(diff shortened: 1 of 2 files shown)


bossie:
An idempotency key allows multiple calls of the same request (= retries) to result in only 1 update. The `/addedvalue` API is called for each unique process ID in the context of 1 process graph; this means n separate requests, each with its unique idempotency key because each one can be retried in isolation. An idempotency key of only request ID/job ID is therefore not adequate.

PR #1678, opened 2026-05-22
job-lifecycle: improve reliability

pvbouwel:
Lower changes of having races against the API.
Explicitly fail when starting a job that is in a final state. Do not check status of Spark application if it weren't created an API exception would have been thrown.

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -1880,7 +1880,7 @@ class GpsBatchJobs(backend.BatchJobs):
                     pass
                 elif current_status == [JOB_STATUS.CANCELED, JOB_STATUS.ERROR, JOB_STATUS.FINISHED]:
                     raise OpenEOApiException(
-                        code="Unsupported job transition",
+                        code="JobTransitionUnsupported",
                         message=f"Backend does not support start job for final job state {current_status}",
                         status_code=400,
                     )
@@ -1890,7 +1890,7 @@ class GpsBatchJobs(backend.BatchJobs):
                         f"GpsBatchJobs._start_job: ill-defined job status transition from {current_status!r} to {JOB_STATUS.CREATED!r}"
                     )
                     raise OpenEOApiException(
-                        code="Not implemented job transition",
+                        code="JobTransitionNotImplemented",
                         message=f"Backend does not implement start job for state {current_status}",
                         status_code=501,
                     )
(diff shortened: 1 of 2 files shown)


JeroenVerstraelen (approved):
LGTM.

openeogeotrellis/backend.py
```diff
@@ -1878,15 +1878,22 @@ def _start_job(self, job_id: str, user: User, get_vault_token: Callable[[str], s
                     return
                 elif current_status == JOB_STATUS.CREATED:
                     pass
+                elif current_status == [JOB_STATUS.CANCELED, JOB_STATUS.ERROR, JOB_STATUS.FINISHED]:
+                    raise OpenEOApiException(
+                        code="Unsupported job transition",
```
soxofaan: code in openEO errors should be CamelCaseWithoutSpaces

openeogeotrellis/backend.py
```diff
@@ -1878,15 +1878,22 @@ def _start_job(self, job_id: str, user: User, get_vault_token: Callable[[str], s
...
-                    logger.warning(
+                    # Another state did not exist at time of having this code so should fail to be safe
+                    logger.error(
                         f"GpsBatchJobs._start_job: ill-defined job status transition from {current_status!r} to {JOB_STATUS.CREATED!r}"
                     )
-                    # TODO: do we really want to reset the application id?
-                    dbl_registry.set_application_id(job_id=job_id, user_id=user_id, application_id=None)
-                    dbl_registry.set_status(job_id=job_id, user_id=user_id, status=JOB_STATUS.CREATED)
+                    raise OpenEOApiException(
+                        code="Not implemented job transition",
```
soxofaan: same

openeogeotrellis/backend.py
soxofaan: for inspiration or alignment, see https://github.com/Open-EO/openeo-api/blob/master/errors.json

openeogeotrellis/backend.py
```diff
@@ -2157,12 +2164,19 @@ def as_boolean_arg(job_option_key: str, default_value: str) -> str:
...
             )
 
             with self._double_job_registry as dbl_registry:
                 try:
+                    latest_job_status = dbl_registry.get_job(job_id=job_id, user_id=user_id)["status"]
+                    if latest_job_status != JOB_STATUS.CREATED:
+                        # While we can again check all different job states the only valid state to proceed is CREATED
+                        # in other cases a concurrent event caused a job transition. Given that at the start of this
+                        # method the state was valid it is best to just NOOP and communicate the job is started.
+                        log.info(f"Job already in state {latest_job_status} ")
```
soxofaan:
```suggestion
                        log.warning(f"Job start requested, but already in state {latest_job_status} ")
```

pvbouwel: Will change the wording but will keep it as info because if there is a warning or error required it will be emitted by the strict check.

soxofaan: FYI: I suggested to change to warning because this is a situation that should not happen from normal usage. If this happens it's a sign of improper usage from client side, or  backend side being under capacity

PR #1693, opened 2026-06-03
s3proxy: force sigv4 signatures for proxy clients

pvbouwel:
Make sure clients do not fallback to legacy signing algorithms leading to hard to troubleshoot bugs. The more modern sigv4 is safer and works well with the S3proxy.


EmileSonneveld:
Does this only change landsat bucket acces? looks safe enough to me

EmileSonneveld:
@copilot review without making changes

PR #1704, opened 2026-06-08
Make GDAL env variables more dynamic

pvbouwel:
This is a draft currently adding test coverage which makes sure GDAL config is in the rendered resulting template.


pvbouwel:
@JeroenVerstraelen just an FYI I do not anticipate any more changes on this PR. I have been able to perform some end-to-end testing which had good results

PR #1710, opened 2026-06-12
be lenient when encountering unknown process id when trying to start job

jdries:


jdries:
@copilot  merge in changes from master branch

Issue #1618, opened 2026-03-25
Serialize/reuse ItemCollection


soxofaan:
spin-off from https://github.com/Open-EO/openeo-python-driver/issues/406:

>  Global extent alignment is not quite fine yet, consider ...:
> - for all load_stac/load_collection cases, serialize queried products to an itemcollection
> 
> ...
> Note that if we build such a static stac collection or itemcollection, we could also improve the 'derived-from' links to rather point to this static catalog, reducing the dependency on the actual on line catalog?
> See: https://github.com/Open-EO/openeo-geopyspark-driver/issues/824

soxofaan:
Merged initial implementation https://github.com/Open-EO/openeo-geopyspark-driver/pull/1658

Serialized item collection is currently added as link to job result metadata with
```json
    "rel": "experimental-derived-from-stac-item-collection"
```

this probably has to be changed later to fully replace existing `"rel": "derived_from"` links, but I didn't want to flip that switch yet (as it will break too much CI and integration tests )

jdries:
@soxofaan  the idea is that you enable this when the STAC 1.1 output is requested.
That way, you can safely enable it without worrying about breaking tests or existing workflows.

soxofaan:
status update: I want to set up an integration test for this, but at the moment I can't even test manually as all dev deployments are broken or stuck at an too old image.

soxofaan:
picking this one up again after being put in the freezer to work on CLMS onboarding https://github.com/eu-cdse/openeo-cdse-infra/issues/686

soxofaan:
ok I finally got something that might work (unit tests are green), but I'm hesitant to merge in master now because we are struggling to get current master live and I don't want to risk making that struggle even harder

soxofaan:
decided to merge it now, with 4ba0ad8c76e5547267ce294909f62e799dda7dbc

soxofaan:
did a quick test on CDSE dev and it seems to work there
will try to set up an integration test when I find some time between all operational issues

soxofaan:
added integration test for CDSE now

soxofaan:
also pushed (same) integration test on terrascope
(unfortunately I could verify some details due to lack of a working dev deploy)

soxofaan:
time to close this ticket

the next steps (cleanup of previous implemenation) is for https://github.com/Open-EO/openeo-geopyspark-driver/issues/1490

PR #1712, opened 2026-06-12
fix item-collection download  on CDSE

soxofaan:
Under #1618, the item-collection with derived-from info was added, and can be downloaded on Terrascope

but it doesn't work yet on kube/CDSE

this (draft) PR aims to fix that


openeogeotrellis/deploy/batch_job_metadata.py
```diff
@@ -442,7 +444,7 @@ def _get_tracker_metadata(tracker_id: str = "", *, omit_derived_from_links: bool
         def _as_auxiliary_links(auxiliary_files) -> List[dict]:
             return [
                 {
-                    "href": str(auxiliary_file.getPath()),
+                    "href": f"openeo-job-local://{auxiliary_file.getPath()}",
```
soxofaan: revert
don't bother trying to improve the situation here, this is turning in a can of worms

openeogeotrellis/util/metadata.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Callable, Optional
+
+
+class MetaDataConvertor:
```
soxofaan: simplify again to a single function for now, to not overkill this PR?

openeogeotrellis/deploy/batch_job_metadata.py
```diff
@@ -470,4 +471,18 @@ def _as_auxiliary_links(auxiliary_files) -> List[dict]:
             usage["input_pixel"]["value"] += sar_backscatter_inputpixels / (1024 * 1024)
         else:
             usage["input_pixel"] = {"value": sar_backscatter_inputpixels / (1024 * 1024), "unit": "mega-pixel"}
-    return dict_no_none(usage=usage or None, links=all_links, auxiliary_links=auxiliary_links or None)
+    return dict_no_none(usage=usage or None, links=all_links, auxiliary_links=auxiliary_links or [])
```
soxofaan: is this change to `auxiliary_links or []` still necessary at this point?

openeogeotrellis/config/config.py
```diff
@@ -292,10 +292,17 @@ class GpsBackendConfig(OpenEoBackendConfig):
         "https://epod-master1.vgt.vito.be:8090",
     )
 
+    batch_job_work_dir_root: str = os.environ.get(
```
soxofaan, line 296: "work_dir_root" sounds bit redundant at first. clarify that it is actually the root under which to put per job work dirs

soxofaan:
- [x] bump version
- [x] add changelog entry

Issue #507, opened 2023-09-11
merge_cubes: invalid spatial_extent merging


soxofaan:
came up when trying to compile a small poc for cross-backend processing (https://github.com/Open-EO/openeo-aggregator/issues/115):
`merge_cubes` only seems to consider the spatial extent of the first cube, ignoring the second.

e.g.

```python
cube1 = connection.load_collection(
    "TERRASCOPE_S2_TOC_V2",
    temporal_extent=["2023-08-20", "2023-09-01"],
    spatial_extent={"west": 3.00, "south": 51.00, "east": 3.10, "north": 51.10},
    bands=["B02"],
 )
cube2 = connection.load_collection(
    "TERRASCOPE_S2_TOC_V2",
    temporal_extent=["2023-08-20", "2023-09-01"],
    spatial_extent={"west": 3.05, "south": 51.05, "east": 3.15, "north": 51.15},
    bands=["B02"],
)

cube = cube1.merge_cubes(cube2, overlap_resolver="max")
res = cube.save_result(format="netCDF")
job = res.execute_batch()
```

Result only covers `{"west": 3.00, "south": 51.00, "east": 3.10, "north": 51.10}` and does not extend to `{..., "east": 3.15, "north": 51.15}`

When switching both cubes (`cube2.merge_cubes(cube1, overlap_resolver="max")`, the result only covers `{"west": 3.05, "south": 51.05, "east": 3.15, "north": 51.15}`

(not sure yet if this is an issue in openeo-geopyspark-driver or openeo-python-driver, just reporting here for now)

soxofaan:
(FYI, just tried to verify:: merging of cubes where only temporal extent is different seems to work properly)

ElienVandermaesenVITO:
The extent that was used to compute the merge process was the extent of the left cube, the metadata does get changed at the end. 
I now changed it that the cubes both have the dimensions of the combined extent and layout before it merged and added test that comfirmed that this fixes the problem.

If I test it now with the initial code and job option ` "openeo-jar-path": "https://artifactory.vgt.vito.be/artifactory/libs-release-public/org/openeo/geotrellis-extensions/PR-784/geotrellis-extensions-PR-784.jar"` it still only saves the left cube.
There is some logging in the `save results` that indeed the extent that is save is the left cube extent. Not sure why it happens, because the netcdf has the merged extent

ElienVandermaesenVITO:
[...]

PR #1714, opened 2026-06-15
issue #507 update metadata geopysparkdatacube

ElienVandermaesenVITO:
#507


soxofaan:
I pushed some finetuning/simplification:
- BoundingBox has a "from_dict()" factory which takes away a lot of boilerplate
- "union_bbox" with west, south, ... arguments also felt heavy on boilerplate.  The "from_dict()" simplification also made the remaining stuff pretty compact anyway and eliminated the urge to encapsulate it in a method at the moment.

let's see what the unit tests say

PR #1717, opened 2026-06-18
Integrate sentinel5p

EmileSonneveld:
Any new Sentinel-5 functionallity will be hidden by default so can't cause regressions.

Some changes that are minor where made too:
- openeogeotrellis/layercatalog.py
Added an new clause: layer_source_type == "file-s5p"

- tests/layercatalog.json
Added SENTINEL3_SLSTR_L2_LST for better testing.

- tests/data_collections/test_sentinel3.py
Fixed a test that is disabled by default: test_sentinel3_slstr_l2_lst


EmileSonneveld:
The failing test `test_load_stac.test_world_oom` is also failing on master, so not related to the code changes here.

PR #1721, opened 2026-06-22
Ignore https://collections.eurodatacube.com/ certificate expired. It …

EmileSonneveld:
…is already covered by integration tests.

@copilot review


soxofaan:
I would just remove these collections. The unit tests should not be making http requests to real external resources

EmileSonneveld:
problem avoided differently on master branch

PR #1722, opened 2026-06-22
Allow catalogue json

EmileSonneveld:
@copilot review


EmileSonneveld:
`test_world_oom` failed for the same reason on master too, so merged.

PR #1726, opened 2026-06-25
Keep Landsat bands with special role types. 

EmileSonneveld:
https://github.com/eu-cdse/openeo-cdse-infra/issues/1198
@bossie I think you added the previous role exceptions. Is this one ok?


soxofaan:
I think we should avoid hardcoding collection ids in the core logic. Can't this be solved through layercatalog configs?
Or just add these roles to the existing allow-list unconditionally

EmileSonneveld:
Changed to have "custom_data_roles" in layer catalog. When specified, it replaces the roles whitelist.

openeogeotrellis/load_stac.py
```diff
@@ -1392,8 +1393,9 @@ class ItemCollection:
 
     # TODO: leverage pystac.ItemCollection in some way ?
 
-    def __init__(self, items: List[pystac.Item]):
+    def __init__(self, items: List[pystac.Item], *, custom_data_roles: Optional[Set[str]] = None):
```
soxofaan, line 1396: as discussed, it feels conceptually wrong  to make "custom_data_roles" an aspect of the ItemCollection constructor. I think this should just be an aspect at the time you collect assets, e.g. an argument to the `iter_items_with_band_assets` method

soxofaan (commented):
this PR is being overruled by https://github.com/Open-EO/openeo-geopyspark-driver/pull/1731

but just putting this conceptual note here

Issue #1737, opened 2026-07-02
filePath invalid with s3 scheme


ElienVandermaesenVITO:
Original post: https://teams.microsoft.com/l/message/19:2941a270bf8e48a2a8e8a23975051c11@thread.skype/1782980659906?tenantId=9e2777ed-8237-4ab9-9278-2c144d6f6da3&groupId=8c9c739d-2544-4def-8cd4-b65970551b70&parentMessageId=1782980659906&teamName=Unit%20TAP&channelName=openEO-users&createdTime=1782980659906

The cause is probably because the `output_dir` is not in the assets here:
https://github.com/Open-EO/openeo-python-driver/blob/ce77ac84bfffdca3feb7aba450c6a7f1609c07a5/openeo_driver/views.py#L1253-L1259

The `output_dir` should be set here: https://github.com/Open-EO/openeo-geopyspark-driver/blob/3d8e0e5b18e08977c673eb01d874aef04bcbe985/openeogeotrellis/backend.py#L2555 , but isn't because it starts with s3

soxofaan:
> Original post: https://teams.micr....

some details from there

job result metadata has asset (`b5a493b6-50b3-4b6c-8a75-3274e7e5eec1_openEO`) with href like
> https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-2607020759334106be8a957ca485260f/results/assets/Zm...A5/c9...c4/s3://openeo-data-prod-otc-waw4-1/batch_jobs/j-2607020759334106be8a957ca485260f/openEO.tif?expires=1783669484

that whole `s3://...` part makes no sense there in the produced href

PR #1738, opened 2026-07-02
issue remove check for s3:// in href for setting output_dir

ElienVandermaesenVITO:
#1737

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1098,9 +1098,55 @@ class TestGpsBatchJobs:
         assert result_metadata.items == {
             "b9510f20-92a0-4947-a4b6-a6a934a0015e": {
                 "id": "b9510f20-92a0-4947-a4b6-a6a934a0015e",
-                "assets": {"openEO": DictSubSet({
-                    "href": "s3://bucket/path/to/openEO.tif",
-                })},
+                "assets": {"openEO": {"href": "s3://bucket/path/to/openEO.tif"}},
+            }
+        }
+
+
+
+    def test_output_dir_s3_results_metadata_uri(
+        self,
+        kube_no_zk,
+        backend_implementation,
+        job_registry,
+        mock_non_swift_s3_bucket,
+    ):
+        self._create_dummy_batch_job(backend_implementation, self._dummy_user)
+
+        job_id, job = next(iter(job_registry.db.items()))
+
+        job_metadata_json_key = f"batch_jobs/{job_id}/job_metadata.json"
+        mock_non_swift_s3_bucket.put_object(
+            Key=job_metadata_json_key,
+            Body=json.dumps(
+                {
+                    "items": [
+                        {
+                            "id": "b9510f20-92a0-4947-a4b6-a6a934a0015e",
+                            "assets": {"openEO": {"href": f"s3://openeo-data-prod-otc-waw4-1/batch_jobs/{job_id}/openEO.tif"}},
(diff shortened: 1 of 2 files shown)


ElienVandermaesenVITO:
I removed the `not asset["href"].startswith("s3://"):` from https://github.com/Open-EO/openeo-geopyspark-driver/blob/3d8e0e5b18e08977c673eb01d874aef04bcbe985/openeogeotrellis/backend.py#L2555 but I'm not certain the job_dir is the right one. It need to be something like `s3://openeo-data-prod-otc-waw4-1/batch_jobs/j-2607020759334106be8a957ca485260f`

tests/test_backend.py
```diff
@@ -1097,7 +1098,9 @@ def test_getters_read_from_s3_results_metadata_uri(
         assert result_metadata.items == {
             "b9510f20-92a0-4947-a4b6-a6a934a0015e": {
                 "id": "b9510f20-92a0-4947-a4b6-a6a934a0015e",
-                "assets": {"openEO": {"href": "s3://bucket/path/to/openEO.tif"}},
+                "assets": {"openEO": DictSubSet({
+                    "href": "s3://bucket/path/to/openEO.tif",
```
soxofaan: why not make it explicit what the output_dir is in this test?

soxofaan: in the failed test I see
```
                'openEO': {
                    'href': 's3://bucket/path/to/openEO.tif',
  +                 'output_dir': '/batch_jobs/j-260702134802413cb2c8fcf711657072',
                },
```

and you know the job id at that point, so you could just build the expected output dir

ElienVandermaesenVITO: I did not make it explicit because I thought it was not relevant to the test, but it is also a good option.

ElienVandermaesenVITO: But do you know the `/openeo-data-prod-otc-waw4-1` part, because there are multiple options?

soxofaan: > But do you know the /openeo-data-prod-otc-waw4-1 part, because there are multiple options?

that's not relevant for this PR, right? Or am I misunderstanding?

ElienVandermaesenVITO: The href is something like: 
`https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-2607020759334106be8a957ca485260f/results/assets/Zm...A5/c9...c4/s3://openeo-data-prod-otc-waw4-1/batch_jobs/j-2607020759334106be8a957ca485260f/openEO.tif?expires=1783669484`, 
what you want is 
`https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-2607020759334106be8a957ca485260f/results/assets/Zm...A5/c9...c4/openEO.tif?expires=1783669484 `
That means that `s3://openeo-data-prod-otc-waw4-1/batch_jobs/j-2607020759334106be8a957ca485260f` is too much, this has to be put into the output_dir and is filtered out later. So `/openeo-data-prod-otc-waw4-1` is also relevant here.

soxofaan: yes, but I think that "prefix" is going to be removed automatically by the current "commonpath/relpath" logic at  https://github.com/Open-EO/openeo-python-driver/blob/ce77ac84bfffdca3feb7aba450c6a7f1609c07a5/openeo_driver/views.py#L1253-L1257
so that's why I don't think it's relevant for this particular PR

Issue #1351, opened 2025-09-16
S5P: other gasses


JeroenVerstraelen:

EmileSonneveld:
@manugv What other gases?
Can we copy the list from the bands here? https://openeo.dataspace.copernicus.eu/openeo/collections/SENTINEL_5P_L2
```
"CO",
"HCHO",
"NO2",
"O3",
"SO2",
"CH4",
"AER_AI_340_380",
"AER_AI_354_388",
"CLOUD_BASE_PRESSURE",
"CLOUD_TOP_PRESSURE",
"CLOUD_BASE_HEIGHT",
"CLOUD_TOP_HEIGHT",
"CLOUD_OPTICAL_THICKNESS",
"CLOUD_FRACTION",
```

manugv:
Yes, you have covered almost all of them. Initially, we only release all gases support which is `CO, CH4, NO2, SO2, HCHO, O3`. Then later, we can add Aerosol index and cloud variables.

manugv:
In this ticket on other gases are `SO2, HCHO, O3`.

EmileSonneveld:
SENTINEL5P_L2_SO2, SENTINEL5P_L2_HCHO and SENTINEL5P_L2_O3 on CDSE dev and have integration tests

PR #1744, opened 2026-07-07
S5p other gases

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1351

### DIFF
--- a/openeogeotrellis/collections/sentinel5p_functions.py
+++ b/openeogeotrellis/collections/sentinel5p_functions.py
@@ -53,6 +53,30 @@ all_gases = {
         "DEFAULT_BANDS": ["methane_mixing_ratio_bias_corrected"],
         "FILTER_VALUE": 0.5,  # default filter value for CH4 as per documentation
     },
+    # SO2 gas variables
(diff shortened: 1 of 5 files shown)


openeogeotrellis/collections/sentinel5p_functions.py
```diff
@@ -122,9 +146,15 @@ def load_data_from_file(
         # if there is valid data, get the pixel indices representing the spatial extents
         # Load time for each row
         var_path = variable_loc_in_file.get("delta_time")
+        delta_time_raw = f[var_path][0]
+        if delta_time_raw.ndim == 2:
+            # Some gas products (e.g. SO2, HCHO, O3) store delta_time per ground pixel
```
EmileSonneveld, line 166: @manugv Is this correct? It was written by copilot

EmileSonneveld:
@copilot review without making code changes

manugv:
yes, that is one way of doing it. Another way is to get rid of all these variable specifications as the informations can also be found the files. This makes the code general and can load any sentinel 5p data.

PR #1774, opened 2026-08-13
Helper signing against internal or public proxy

JeroenVerstraelen:
https://github.com/eu-cdse/openeo-cdse-infra/issues/1224

### DIFF
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -102,6 +102,7 @@ from openeogeotrellis.integrations.kubernetes import (
 )
 from openeogeotrellis.integrations.s3proxy.asset_urls import PresignedS3AssetUrls
 from openeogeotrellis.integrations.s3proxy.s3_shared_context import presign_s3_urls_for_internal_usage
+from openeogeotrellis.integrations.s3proxy.s3_uris import get_bucket_key_from_uri
 from openeogeotrellis.integrations.stac import ResilientStacIO
 from openeogeotrellis.integrations.traefik import Traefik
 from openeogeotrellis.integrations.yarn_jobrunner import YARNBatchJobRunner
@@ -2651,7 +2652,7 @@ class GpsBatchJobs(backend.BatchJobs):
             if uri_parts.scheme == "file":
                 return try_get_results_metadata_from_disk(uri_parts.path)
             elif uri_parts.scheme == "s3":
-                bucket, key = PresignedS3AssetUrls.get_bucket_key_from_uri(results_metadata_uri)
+                bucket, key = get_bucket_key_from_uri(results_metadata_uri)
                 return try_get_results_metadata_from_object_storage(key, bucket)
             else:
                 raise NotImplementedError(results_metadata_uri)
--- a/openeogeotrellis/backend.py
+++ b/openeogeotrellis/backend.py
@@ -56,6 +56,7 @@ from openeo_driver.dry_run import SourceConstraint, DryRunDataCube, DryRunDataTr
 from openeo_driver.errors import (InternalException, JobNotFinishedException, OpenEOApiException,
                                   ServiceUnsupportedException,
                                   ProcessParameterInvalidException, ProcessUnsupportedException, )
+from openeo_driver.integrations.s3.presigned_url import create_presigned_url
 from openeo_driver.jobregistry import (DEPENDENCY_STATUS, JOB_STATUS, ElasticJobRegistry, PARTIAL_JOB_STATUS,
                                        get_ejr_credentials_from_env)
 from openeo_driver.ProcessGraphDeserializer import ENV_FINAL_RESULT, ENV_SAVE_RESULT, ConcreteProcessing, \
@@ -101,6 +102,7 @@ from openeogeotrellis.integrations.kubernetes import (
     ensure_kubernetes_config,
 )
 from openeogeotrellis.integrations.s3proxy.asset_urls import PresignedS3AssetUrls
+from openeogeotrellis.integrations.s3proxy.s3_shared_context import get_proxy_s3_client_for_job
 from openeogeotrellis.integrations.stac import ResilientStacIO
 from openeogeotrellis.integrations.traefik import Traefik
 from openeogeotrellis.integrations.yarn_jobrunner import YARNBatchJobRunner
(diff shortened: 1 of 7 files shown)


openeogeotrellis/backend.py
```diff
@@ -1522,6 +1524,7 @@ def get_elastic_job_registry(
 
 
 class GpsBatchJobs(backend.BatchJobs):
+    _UDF_DEPENDENCY_PRESIGNED_URL_EXPIRATION = 24 * 3600
```
pvbouwel: nit: seems to belong in config

openeogeotrellis/backend.py
```diff
@@ -2323,6 +2331,77 @@ def as_boolean_arg(job_option_key: str, default_value: str) -> str:
...
+            )
+        if "#" in s3_uri:
+            s3_uri_without_alias, alias = s3_uri.split("#", 1)
+            if not alias:
+                raise OpenEOApiException(
+                    status_code=400,
+                    message=f"Invalid S3 UDF dependency file {s3_uri!r}: expected non-empty alias after '#'.",
+                )
+            return s3_uri_without_alias, alias
+        raise OpenEOApiException(
```
pvbouwel: An alias is not mandatory for the spark spec and I think this stricter enforcement would be a breaking change

openeogeotrellis/backend.py
```diff
@@ -2323,6 +2331,77 @@ def as_boolean_arg(job_option_key: str, default_value: str) -> str:
...
+                    message=f"Generated invalid presigned url {presigned_url!r} for {s3_uri_without_alias} job_id={job_id} user={user_id}",
+                )
+            if parsed_presigned_url.fragment:
+                raise OpenEOApiException(
+                    status_code=400,
+                    message=f"Generated presigned url unexpectedly contains fragment: {presigned_url!r} for {s3_uri_without_alias} job_id={job_id} user={user_id}",
+                )
+
+            resolved_dependency_file = f"{presigned_url}#{alias}"
+            parsed_resolved_dependency_file = urlparse(resolved_dependency_file)
```
pvbouwel: Not sure why this alias part is checked at runtime, is there a particular scenario where you expect this to happen that you'd want to put code in place? Having a test case in your unittests seems sufficient.

openeogeotrellis/integrations/s3proxy/s3_shared_context.py
```diff
@@ -38,7 +47,10 @@ def _get_proxy_s3_client_for_job(bucket: str, job_id: str, user_id) -> S3Client:
     bucket_details = BucketDetails.from_name(bucket)
     region_name = bucket_details.region
     try:
-        endpoint_url = get_backend_config().s3_region_proxy_endpoints[region_name]
+        if internal:
+            endpoint_url = os.environ[AWSConfig.S3PROXY_S3_ENDPOINT_URL]
```
pvbouwel, line 66: Here likely need handling if region is the default _REGION_UNKNOWN because that would mean the bucket is unknown (no workspace config in place that provides region information for the bucket).

pvbouwel (changes requested):
Added inline comments as well but it feels wrong to put this code in backend.py. The current implementation acts like it is exclusive for UDF dependencies (given the naming of methods) but it feels more generic. I believe it makes more sense to use something like `presign_s3_urls_for_internal_usage` and have it live in `openeogeotrellis.integrations.s3proxy.s3_shared_context`. Testing can then as well be done in that scope and avoid the extensive mocking. If you just use the fixtures `moto_server`,`mock_s3_client` and `mock_sts_client` then there should not be actual API calls and moto mocking should even make s3 urls that work. So you can have a test fixture that create a bucket with an object in there and make sure it becomes a downloadable presigned url.  Then you likely need to adapt the fixture _dynamic_overrides to mock the env var AWSConfig.S3PROXY_S3_ENDPOINT_URL to moto_server_address.

openeogeotrellis/integrations/s3proxy/s3_shared_context.py
```diff
@@ -61,3 +70,78 @@ def _get_proxy_s3_client_for_job(bucket: str, job_id: str, user_id: str, interna
...
+    from urllib.parse import urlparse
+
+    resolved = []
+    for uri in s3_uris:
+        if not uri.startswith("s3://"):
+            resolved.append(uri)
+            continue
+
+        # Avoid circular import: asset_urls imports from this module
+        from openeogeotrellis.integrations.s3proxy.asset_urls import PresignedS3AssetUrls
```
pvbouwel: Very ugly local import to avoid a circular dependency. In this case get_bucket_key_from_uri is a static method which can be a regular function somewhere in another module (e.g. integrations/s3proxy/s3_uris.py)

openeogeotrellis/integrations/s3proxy/s3_shared_context.py
```diff
@@ -46,6 +51,10 @@ def _get_proxy_s3_client_for_job(bucket: str, job_id: str, user_id: str, interna
     """
     bucket_details = BucketDetails.from_name(bucket)
     region_name = bucket_details.region
+    if region_name == _REGION_UNKNOWN:
```
pvbouwel: I did not mean literaly comparing with this because _REGION_UNKNOWN is an internal var. So you would either need to add a `has_valid_region` helper or use the is_workspace_bucket (all workspace buckets have regions)

pvbouwel (commented):
Some minor comments

JeroenVerstraelen:
@pvbouwel should be ready for review again.

openeogeotrellis/integrations/s3proxy/s3_uris.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def split_s3_uri_and_alias(s3_uri: str) -> Tuple[str, Optional[str]]:
+    """Split an S3 URI into the URI without alias and an optional alias fragment.
+
+    Returns (s3_uri_without_alias, alias) where alias may be None if no '#' is present.
+    Raises OpenEOApiException if the URI has more than one '#' or an empty alias.
+    """
+    if s3_uri.count("#") > 1:
+        raise OpenEOApiException(
```
pvbouwel, line 32: nit: Seems a bit weird to throw OpenEOApiExceptions from utility functions. Generally this would be a ValueError which should be caught at the layer that is OpenEO aware (e.g. backend.py) and where it should raise an OpenEOApiException from the error that was thrown.

Because now you assume s3_uri comes from something the user provided (which is the case for this use case) and therefore it is a bad request but if this utility function gets used from internal code and remains uncaught it should be a 5xx error.

pvbouwel (changes requested):
Tests need to be fixed and added 1 nit but that one is not really blocking

PR #1779, opened 2026-08-17
fix: do not change job_options after parsing

pvbouwel:


JeroenVerstraelen (approved):
job_options is still used down the line by a few other functions, so this change could be dangerous. But I checked and image_name is not used anywhere. So it suffices as a quick fix.

JeroenVerstraelen (approved):

Issue #1252, opened 2025-06-16
[EPIC] Direct support for Sentinel 5P L2 products


jdries:
Add support to read sentinel 5P L2 products directly.
We may want to immediately start from the zarr format, avoiding to do the implementation two times.

JeroenVerstraelen:
Hard Deadline: CDSE User review early September.

EmileSonneveld:
@manugv What would be the best option?

Multiple collections:
```
SENTINEL5P_L2_CO
["carbonmonoxide_total_column", "carbonmonoxide_total_column_corrected", "qa_value"]

SENTINEL5P_L2_NO2
["nitrogendioxide_tropospheric_column", "qa_value"]

SENTINEL5P_L2_CH4
["methane_mixing_ratio", "methane_mixing_ratio_bias_corrected", "qa_value"]
```
Single collection (Just like sentinelhub example https://openeo.dataspace.copernicus.eu/openeo/collections/SENTINEL_5P_L2 )
```
SENTINEL5P_L2
[
  "carbonmonoxide_total_column", "carbonmonoxide_total_column_corrected", "carbonmonoxide_total_column_qa_value",
  "nitrogendioxide_tropospheric_column", "nitrogendioxide_tropospheric_column_+qa_value"
  "methane_mixing_ratio", "methane_mixing_ratio_bias_corrected", "methane_mixing_ratio_qa_value"
]
```
Not sure if prefixing makes sense for the `qa_value` band.

TAdeJong:
@EmileSonneveld: Two cents from my side: Note the different products might each have different conflicting values: the CO product also contains (non-scattering retrieved) methane, the CH4 product also retrieves and contains a CO column. So if not in separate collections, it would need some clear documentation what is actually in there.

manugv:
1. We will be doing multiple collections. 
2. The code in the back-end will be the same. Now it only works for certain gases and I will make it general soon.
3. We should further discuss on where default bands are specified in python-client side or in the back-end. I prefer python client side.

EmileSonneveld:
TODO: Notify this forum post when changes hit production: https://forum.dataspace.copernicus.eu/t/quality-factor-for-sentinel-5p-l2-so2-products/3819

JeroenVerstraelen:
Some notes after speaking with Jeroen Dries.

@EmileSonneveld 
[...]

PR #1801, opened 2026-08-27
Fix clamping issue when using nearest interpolation.

manugv:


EmileSonneveld (approved):
All tests pass when having /eodata mounted. LGTM. Not sure if we can merge now, with the code freeze

EmileSonneveld:
Code freeze over. Can merge.
This also allows to skip band names, and puts the default resolution higher.
Related to: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1252

Issue #1804, opened 2026-09-02
kafka dependency?


soxofaan:
there is a [security advisory](https://github.com/advisories/GHSA-2jcm-hq8r-84wx) about the kafka dependency in geopyspark driver

but question is if we still need kafka (as required dependency) at all

soxofaan:
cc @bossie

soxofaan:
related:
- https://github.com/Open-EO/openeo-geopyspark-driver/pull/1803

soxofaan:
kafka usage has been phased out indeed:
- eu-cdse/openeo-cdse-infra#387
- eu-cdse/openeo-cdse-infra#339

soxofaan:
initial attempt to drop kafka and related code:
- https://github.com/Open-EO/openeo-geopyspark-driver/pull/1805

let's see what CI and AI thinks of it

soxofaan:
full removal is apparently more of a rabbit hole than expected

so started with a much easier first step: make kafka an optional dependency:
- #1806

soxofaan:
merged #1805   in a80b5d14a4b3cd9a720b431377741f861c8c1b51

PR #1805, opened 2026-09-02
Attempt to remove all traces of kafka and "async_task" #1804

soxofaan:
all this code should be dead/unreachable since eu-cdse/openeo-cdse-infra#387 and eu-cdse/openeo-cdse-infra#339


soxofaan:
@copilot this is attempt to remove unreachable code, related to the now used kafka and the "async_task" feature.
Can you check the code change (no need to run tests or do fixup commits)  if nothing is forgotten or removed to aggressively?

soxofaan:
currently failing tests:
```
tests.test_views.TestSentinelHubBatchJobs.test_start_large_extent[1.0.0]
tests.test_views.TestSentinelHubBatchJobs.test_start_large_extent_with_polarization_based_on_bands[1.0.0]
tests.test_views.TestSentinelHubBatchJobs.test_download_large_extent[1.0.0]
tests.test_views.TestSentinelHubBatchJobs.test_download_pending_signed_job_results[1.0.0]
tests.test_views.TestSentinelHubBatchJobs.test_download_pending_own_job_results[1.0.0]
```

but I guess this is expected as these tests cover the "async_task" feature.

@bossie   is it safe to also just remove the whole TestSentinelHubBatchJobs test class?

bossie:
The failing tests are indeed because, with `async_task` gone, it tries to start an OpenEO batch job immediately rather than schedule SHub batch processes.

The remaining tests in `TestSentinelHubBatchJobs` are probably covered elsewhere so yes, it can go entirely.

Removing this class will show that `openeogeotrellis.backend.GpsBatchJobs.poll_job_dependencies` is no longer called so that can go as well.

soxofaan:
ok thanks,

removing `TestSentinelHubBatchJobs` and `poll_job_dependencies` seems to work based on unit test results

further progress (e.g. merging this) is a bit blocked by current code freeze however

PR #1806, opened 2026-09-02
Make kafka an optional dependency for now #1804

soxofaan:
first step to remove kafka dependency (#1804) : make it optional

(more simple alternative for full removal from #1805  )


soxofaan:
overruled by #1805 (now merged)

Issue #1819, opened 2026-09-16
Sentinel5p artifact on 2023-06-29


EmileSonneveld:
<img width="652" height="271" alt="Image" src="https://github.com/user-attachments/assets/2cb89d05-ec95-4eec-bf8f-6a9115b429cf" />

```json
{
  "process_graph": {
    "loadcollection1": {
      "process_id": "load_collection",
      "arguments": {
        "bands": ["aerosol_index_354_388"],
        "id": "SENTINEL5P_L2_AER_AI",
        "spatial_extent": {"east": 8, "north": 54, "south": 50, "west": 3},
        "temporal_extent": ["2023-06-29T13:00:00.000000Z", "2023-06-29T23:00:00.000000Z"]
      },
      "result": true
    }
  },
  "parameters": []
}
```
@manugv This might a masking issue

EmileSonneveld:
It seems like the masking polygon is invalid:
<img width="1689" height="292" alt="Image" src="https://github.com/user-attachments/assets/7d1c365c-58b1-455a-a9aa-58341692e77d" />

EmileSonneveld:
@manugv , as you mentioned, ignoring data above 85deg seems to fix already one issue:
Red: no ignore
Green: do ignore above 85deg
<img width="886" height="428" alt="Image" src="https://github.com/user-attachments/assets/2f486f68-ad97-4727-9b97-bdd8510c5b18" />

EmileSonneveld:
There is also some anti-meridian wrapping at play. Added `test_data_loading_with_complex_bounding_box_02` that reproduces this: 
<img width="1596" height="400" alt="Image" src="https://github.com/user-attachments/assets/fcfe5785-8d7d-425c-ada8-4e73e4685097" />

PR #1824, opened 2026-09-17
Test sentinel5p l2 artifacts

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1819


EmileSonneveld:
@manugv Do you want to take this fix over? There is a polygon that gets wrongly wrapped around the antimeridian. I think it needs to be cut in 2 while it is in the [0-360] range (as opposed to [-180,180] range

manugv:
@EmileSonneveld  If the deadline is end of the week then I can take over. If it is like tomorrow or day after then I prefer you take over.

EmileSonneveld:
Ideally next release, so merged before 1 october

manugv:
I can fix it by Sept 28/29 so I will take it over.

PR #1825, opened 2026-09-18
Return custom_processes in both cases in load_custom_processes, so th…

EmileSonneveld:
…e calling code can check if it was successful.

`test_custom_processes.py` has some obscure errors when running, because `load_custom_processes` fails without letting know


EmileSonneveld:
I tried to solve an unrelated error in this PR. Closed and just did the simple commit here: https://github.com/Open-EO/openeo-geopyspark-driver/commit/848d69b1a99522087afc08359c73419637a2227c

